PR #6, opened 2019-07-02
[mds-compliance] [mds-db] Add support for historical Compliance Snapshots

avatarneil:
This PR adds the ability to specify an `end_date` in a request to the `compliance/snapshot` endpoint, to enable historical lookup. Easy possible extension of this could be adding an optional `start_date` parameter as well, for more accurate reconstruction.

## PR Checklist

## Impacts
- [ ] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [x] Compliance

### DIFF
--- a/packages/mds-compliance/api.ts
+++ b/packages/mds-compliance/api.ts
@@ -145,23 +145,54 @@ function api(app: express.Express): express.Express {
     }
     /* istanbul ignore next */
     function fail(err: Error): void {
-      log.error('readDeviceIds fail', err.stack || err).then(() => {
+      log.error(err.stack || err).then(() => {
         res.status(500).send('server error')
       })
     }
 
-    const start_date = now() - days(365)
-    const end_date = now() + days(365)
+    let start_date = now() - days(365)
     const { policy_uuid } = req.params
+    const { provider_id } = req.query
+    let { end_date } = req.query
 
     if (!isUUID(policy_uuid)) {
       res.status(400).send({ err: 'bad_param' })
+    } else if (end_date) {
+      end_date = parseInt(end_date)
+      start_date = end_date - days(365)
+      try {
+        const policies = await db.readPolicies({ policy_id: policy_uuid, start_date, end_date })
+        const geographies = await db.readGeographies()
+        const deviceIdsWithProvider = await db.readDeviceIds(provider_id)
+        const deviceIds = deviceIdsWithProvider.reduce((acc: UUID[], deviceId) => {
+          return [...acc, deviceId.device_id]
+        }, [])
+        const devices = await cache.readDevices(deviceIds)
+        const deviceMap = devices.reduce((map: { [d: string]: Device }, device) => {
+          /* eslint-disable-next-line no-param-reassign */
+          map[device.device_id] = device
+          return map
+        }, {})
+        const events = await db.readHistoricalEvents({ provider_id, end_date })
+        const filtered_policies: Policy[] = compliance_engine.filterPolicies(policies)
+        const results: (ComplianceResponse | undefined)[] = filtered_policies.map((policy: Policy) =>
+          compliance_engine.processPolicy(policy, events, geographies, deviceMap)
+        )
+        if (!results[0]) {
+          res.status(400).send({ err: 'bad_param' })
+        } else {
+          res.status(200).send(results)
+        }
+      } catch (err) {
+        fail(err)
+      }
     } else {
+      end_date = now() + days(365)
       db.readPolicies({ policy_id: policy_uuid, start_date, end_date })
         .then((policies: Policy[]) => {
           db.readGeographies()
             .then((geographies: Geography[]) => {
-              db.readDeviceIds(req.query.provider_id)
+              db.readDeviceIds(provider_id)
                 .then((deviceRecords: { device_id: UUID; provider_id: UUID }[]) => {
                   const total = deviceRecords.length
                   log.info(`read ${total} deviceIds in /vehicles`)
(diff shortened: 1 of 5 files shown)


packages/mds-db/index.ts
```diff
@@ -537,6 +550,100 @@ async function readEvents(params: ReadEventsQueryParams): Promise<ReadEventsResu
   })
 }
 
+async function readHistoricalEvents(params: ReadHistoricalEventsQueryParams) {
+  const { provider_id: query_provider_id, end_time } = params
+  const client = await getReadOnlyClient()
+  const vals = new SqlVals()
+  const values = vals.values()
+  let sql = `SELECT      e2.provider_id,
```
mdurling: That's an awful lot of SQL ... Not sure I have a recommendation because views and stored procedures are also problematic, but maybe it would be easier to reason about if it was extracted into it's own function. Also, if the entire thing was constructed using template strings, you could put in new lines for readability rather than using `\n`

packages/mds-db/index.ts
```diff
@@ -537,6 +550,100 @@ async function readEvents(params: ReadEventsQueryParams): Promise<ReadEventsResu
...
+FROM
+(
+SELECT      provider_id,
+          device_id,
+          event_type,
+          timestamp,
+          recorded,
+          RANK() OVER (PARTITION BY device_id ORDER BY timestamp DESC) AS rownum
+FROM        events
+WHERE       provider_id IS NOT NULL`
```
mdurling: Can `provider_id` even be `NULL`?

packages/mds-db/index.ts
```diff
@@ -537,6 +550,100 @@ async function readEvents(params: ReadEventsQueryParams): Promise<ReadEventsResu
...
+      timestamp,
+      recorded,
+      lat,
+      lng,
+      speed,
+      heading,
+      accuracy,
+      altitude
+    } = row
+    acc.push({
```
mdurling: In general, I have a preference for immutable array functions. I try to avoid `.push`. It doesn't really matter here, but 
```
return acc.concat({...})
```
or
```
return [...acc, {...}]
```
would be immutable alternatives.

avatarneil:
Test!

avatarneil:
rerun tests

packages/mds-compliance/api.ts
```diff
@@ -145,23 +145,55 @@ function api(app: express.Express): express.Express {
...
+        }, [])
+        const devices = await cache.readDevices(deviceIds)
+        /* eslint-disable-next-line */
+        const deviceMap = devices.reduce((map: {[d: string]: Device}, device) => ((map[device.device_id] = device), map), {})
+        const events = await db.readHistoricalEvents({ provider_id, end_date })
+        const filtered_policies: Policy[] = compliance_engine.filterPolicies(policies)
+        const results: (ComplianceResponse | undefined)[] = filtered_policies.map((policy: Policy) =>
+          compliance_engine.processPolicy(policy, events, geographies, deviceMap)
+        )
+        if (results.length === 0) {
```
janedotx: wat do if all `[undefined, undefined ...]`

packages/mds-compliance/api.ts
```diff
@@ -145,23 +145,55 @@ function api(app: express.Express): express.Express {
...
+      end_date = now() + days(365)
+    }
 
     if (!isUUID(policy_uuid)) {
       res.status(400).send({ err: 'bad_param' })
+    } else if (end_date) {
+      end_date = parseInt(end_date)
+      start_date = end_date - days(365)
+      try {
+        const policies = await db.readPolicies({ policy_id: policy_uuid, start_date, end_date })
```
janedotx: Not a blocker, but I imagine this is going to be a rather slow endpoint with all these reads to the DB. We'll need to change this to hit the cache, or figure out a more efficient way to query the DB, or both.

packages/mds-db/index.ts
janedotx: No, it cannot.

packages/mds-compliance/api.ts
avatarneil: It is pretty slow, yeah. Basically everything in `compliance` is mega slow -- even using cache. There is no way to construct historical state from cache, we _must_ use the db. If you look below, you can see how we compute compliance using cache.

packages/mds-compliance/api.ts
avatarneil: This should probably be a check for `if (results[0])` instead. In this particular case, we're never going to have `policies` be greater than a length of 1 (`policy_id` is provided as a param to the endpoint), so we'll never have more than one item in the `results` list. So the case `[undefined, undefined, undefined ...]` is impossible here.

packages/mds-compliance/api.ts
janedotx: Ah, the inability to use the cache makes sense. I don't know much SQL, but there must be a better way to hit the DB.

avatarneil: That gigantic SQL query is the best our resident SQL master could come up with for reconstructing historical state, I don't think it's going to get much more efficient than this. This endpoint is likely not going to be used in real-time capacities, and more likely by background tasks for aggregating compliance data.

packages/mds-compliance/api.ts
janedotx: I'll approve once that change to the results check comes in.

packages/mds-compliance/api.ts
janedotx: 👍

packages/mds-compliance/api.ts
```diff
@@ -145,23 +145,55 @@ function api(app: express.Express): express.Express {
...
+          return acc
+        }, [])
+        const devices = await cache.readDevices(deviceIds)
+        /* eslint-disable-next-line */
+        const deviceMap = devices.reduce((map: {[d: string]: Device}, device) => ((map[device.device_id] = device), map), {})
+        const events = await db.readHistoricalEvents({ provider_id, end_date })
+        const filtered_policies: Policy[] = compliance_engine.filterPolicies(policies)
+        const results: (ComplianceResponse | undefined)[] = filtered_policies.map((policy: Policy) =>
+          compliance_engine.processPolicy(policy, events, geographies, deviceMap)
+        )
```
janedotx: improve the results check

PR #9, opened 2019-07-03
[mds-*] Removing or converting FIXMEs, or fixing FIXMEs.

janedotx:
## PR Checklist

## Impacts
- [ ] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance


janedotx:
Trello cards made to track various TODOs to help increase visibility of those remaining TODOs.

avatarneil:
Hey Jenkins! Pretty please run tests <3

packages/ladot-service-areas/ladot-service-areas.js
```diff
@@ -3,8 +3,6 @@ const restrictedAreas = require('./restricted-areas')
 const laDacs = require('./la-dacs')
 const veniceSpecOps = require('./venice-special-ops-zone')
 
-// FIXME move into DB eventually
```
avatarneil: Removed because we're deprecating service areas SoonTM?

packages/mds-daily/api.ts
```diff
@@ -24,7 +24,7 @@ import stream from 'mds-stream'
 import { providers, providerName } from 'mds-providers'
 import areas from 'ladot-service-areas'
 import { UUID, VehicleEvent, Telemetry, CountMap, DeviceID } from 'mds'
-import { VEHICLE_EVENTS, VEHICLE_STATUSES, EVENT_STATUS_MAP } from 'mds-enums' // FIXME replace eventually
+import { VEHICLE_EVENTS, VEHICLE_STATUSES, EVENT_STATUS_MAP } from 'mds-enums' // TODO replace eventually
```
avatarneil: I don't see why the enums are something we should replace eventually... Can @karcass weigh in on keeping this comment?

packages/mds-daily/types.ts
```diff
@@ -18,11 +18,12 @@ export interface AgencyApiRequest extends express.Request {
   }
 }
 
+// TODO Extract shared types from mds-agency
 export interface TripsStats {
   single: number
-  singles?: any // FIXME
-  mysteries?: any // FIXME
-  mystery_examples?: any // FIXME
+  singles?: any // TODO
```
avatarneil: This type is duplicated in `mds-agency`, we should move it to the shared `mds.d.ts` file.

packages/mds-db/types.ts
```diff
@@ -50,8 +50,8 @@ export interface StatusChange {
   vehicle_id: string
   vehicle_type: VEHICLE_TYPE
   propulsion_type: PROPULSION_TYPE[]
-  event_type: string // FIXME enum
-  event_type_reason: string // FIXME enum
+  event_type: VEHICLE_STATUS
+  event_type_reason: VEHICLE_EVENT | VEHICLE_REASON
```
avatarneil: Yas, good types 😍

packages/mds-policy/api.ts
```diff
@@ -378,10 +373,10 @@ function api(app: express.Express): express.Express {
     const validation = Joi.validate(policy, policySchema)
     const details = validation.error ? validation.error.details : null
 
-    // FIXME is basically identical to POST policy
+    // TODO is basically identical to POST policy
 
     if (details) {
-      log.info('questionable policy json', details)
+      log.info('policy JSON', details)
```
avatarneil: This is an error condition, and should be treated as such (even though this endpoint will likely be cleaned up a bunch in the near future). Should be Perhaps something like `log.error('Invalid Policy JSON', details)'` as the log message.

packages/mds-policy/api.ts
```diff
@@ -408,7 +403,7 @@ function api(app: express.Express): express.Express {
   /* eslint-enable @typescript-eslint/no-unused-vars */
 
   app.get(pathsFor('/test/initialize'), (req, res) => {
-    // FIXME janky
+    // TODO optimize
```
avatarneil: How's this non-optimal?

packages/mds-provider/api.ts
```diff
@@ -327,15 +320,17 @@ function api(app: express.Express): express.Express {
     return Promise.all(promises)
   }
 
-  // FIXME add pagination?
+  // TODO validate query params,
```
avatarneil: This is a very different TODO than the original FIXME... Can anyone weigh in a bit?

packages/mds-provider/api.ts
```diff
@@ -327,15 +320,17 @@ function api(app: express.Express): express.Express {
...
     log.warn(provider_id ? providerName(provider_id) : 'none', '/trips', JSON.stringify(req.params))
 
-    const { skip, take } = asPagingParams(req.query)
+    const { skip } = asPagingParams(req.query)
     const { start_time, end_time, device_id, newSkool } = req.query
 
-    // FIXME validate start_time, end_time, etc.
+    const PAGE_SIZE = 10 // set low because this is an expensive query.
+
+    const take = Number(env.PAGE_SIZE) || PAGE_SIZE
```
avatarneil: I'm pretty sure forcing `take` to be a PAGE_SIZE set by the constant or an environment variable invalidates the spec -- the client should be able to specify a `take`. If we want to limit how high it can be, this we should still extract the param. Could do something like:
```typescript
const { skip, take: take_param } = asPagingParams(req.query)

const PAGE_SIZE = 10

const take = min(env.PAGE_SIZE, PAGE_SIZE, take_param)
```

packages/mds-utils/index.ts
```diff
@@ -554,7 +554,7 @@ function isStateTransitionValid(eventA: VehicleEvent, eventB: VehicleEvent) {
       }
     case VEHICLE_STATUSES.inactive:
       switch (eventB.event_type) {
-        case VEHICLE_EVENTS.register: // FIXME not totally certain about this one..? Check w/ max
+        case VEHICLE_EVENTS.register:
```
avatarneil: Still gotta check this! @karcass What do you think, do `register` events match to an `inactive` status?

packages/mds-api-helpers/index.ts
```diff
@@ -143,3 +143,20 @@ export const asJsonApiLinks = (req: express.Request, skip: number, take: number,
...
+  error: string
+  error_description: string
+
+  constructor() { 
+    this.error = 'server_error'
+    this.error_description = 'an internal server error has occurred and been logged'
+   }
+}
+
+export const SERVER_ERROR = new ServerError()
```
mdurling: There are a bunch of `Error` classes in `mds-audit/src/exceptions.ts` (including `ServerError`)... now that we have `mds-api-helpers` perhaps those should be moved here and used throughout.

packages/mds-db/index.ts
```diff
@@ -1163,14 +1152,14 @@ async function readTrips(params: ReadEventsQueryParams & { skip: DBVal; take: DB
     if (!isTimestamp(min_end_time)) {
       throw new Error(`invalid min_end_time ${min_end_time}`)
     } else {
-      conditions.push(`min_end_time >= ${vals.add(min_end_time)}`) // FIXME confirm >=
+      conditions.push(`min_end_time >= ${vals.add(min_end_time)}`)
```
mdurling: I'm pretty sure `min_end_time` and `max_end_time` are not actual columns. This is probably a bug waiting to happen. Likely this should be `timestamp >= ${vals.add(min_end_time)}` ... this isn't the only instance of this.

packages/mds-policy/api.ts
mdurling: Could be simply:
```
try {
  const kind = await db.initialize()
  res.status(200).send({ result: '....' })
}
catch (err) {
  await log.error('initialize exception', ex)
  res.status(500).send(SERVER_ERROR)
}
```
}

packages/mds-policy/tsconfig.json
```diff
@@ -4,11 +4,26 @@
     "outDir": "dist"
   },
   "references": [
-    { "path": "../../packages/mds-api-server" },
-    { "path": "../../packages/mds-db" },
-    { "path": "../../packages/mds-enums" },
-    { "path": "../../packages/mds-logger" },
-    { "path": "../../packages/mds-providers" },
-    { "path": "../../packages/mds-utils" }
+    {
```
mdurling: Again with the formatting?

packages/mds-provider/api.ts
mdurling: Provider stuff going to change anyway.
```
const { skip, take } = asPagingParams(req.query)
```
And then pass in:
```
take: Math.min(take, PAGE_SIZE)
```

types/mds.d.ts
```diff
@@ -209,7 +209,6 @@ export type CountMap<T extends {}> = { [P in keyof T]: number }
 // database. This type alias will add the readonly attribute to all properties and also remove undefined as a valid
 // value since the database will never return undefined.
 export type Recorded<T> = Readonly<Required<T>>
-
```
mdurling: Revert this. Spacing increases readability.

packages/ladot-service-areas/ladot-service-areas.js
janedotx: Yah!

packages/mds-daily/types.ts
janedotx: Michael just pushed a change that did this.

packages/mds-policy/api.ts
janedotx: I'm really not sure. @karcass could you weigh in?

packages/mds-provider/api.ts
janedotx: Well it's got pagination in the method body, so it's been added.

packages/mds-policy/tsconfig.json
janedotx: I figured it out. I'm not sure why yours works still, but for me, disabling `formatOnSave` for json and jsonc files prevents the weird formatting.

packages/mds-daily/api.ts
karcass: Stale comment.  Remove.

packages/mds-policy/api.ts
karcass: I don't recall why I wrote that.  It's possible that the code got cleaned up but the comment was not removed.  I don't see any reason to keep the comment.

packages/mds-provider/api.ts
karcass: Sounds like it's ready to be removed.

packages/mds-provider/api.ts
karcass: The client may specify the take, but all of our paging should use "sensible" defaults.

packages/mds-utils/index.ts
karcass: no, they should go to `removed`

packages/mds-api-helpers/index.ts
karcass: yes to this

packages/mds-db/index.ts
janedotx: Yeah, you're right.

janedotx: I just went through the code and that was the only instance I could find of `min_end_time` and `max_end_time` being used instead of `timestamp`, actually.

packages/mds-provider/api.ts
janedotx: Looking at the body, there's param validation actually, so removing this todo.

avatarneil:
@janedotx conflicts

PR #10, opened 2019-07-03
Cleanup eslint errors

mdurling:
Now that we are ~100% typescript we can clean up the ESLint config as we don't need to worry about linting config files and such. Also, resolve all linter warnings and turn off the `only-warn` plugin.


avatarneil:
Mr. Jenkins! Please run tests

mdurling:
Please run tests again

packages/mds-agency/api.ts
```diff
@@ -1154,8 +1165,8 @@ function api(app: express.Express): express.Express {
         return map
       }, eventSeed)
       const telemetrySeed: { [s: string]: Telemetry } = {}
-      const telemetryMap = telemetry.reduce((map, telemetry) => {
-        map[telemetry.device_id] = telemetry
+      const telemetryMap = telemetry.reduce((map, t) => {
+        map[t.device_id] = t
```
avatarneil: Could this be refactored using `Object.assign()`? Something like:
```typescript
const telemetryMap = telemetry.reduce((map, t) => {
   return Object.assign(map, { [t.device_id]: t })
}
```

mdurling: Yes. And in fact, we do this so often that we should create a helper function.

avatarneil:
Mr. Jenkinsssss, bring me a dream (please rerun tests) 
Make this the best PR that I've ever seen!

mdurling:
Run tests with updated Jenkins build steps

PR #15, opened 2019-07-08
[mds-provider] Updated Event Processing

mdurling:
More updates to the provider event processing to include processing db events that predate the provider event stream.

## Impacts
- [x] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,7 +4,7 @@ module.exports = {
     es6: true,
     mocha: true,
   },
-  plugins: ['@typescript-eslint', 'prettier'],
+  plugins: ["import", "prettier", "@typescript-eslint"],
   parser: '@typescript-eslint/parser',
   parserOptions: {
     ecmaVersion: 2018,
@@ -37,7 +37,6 @@ module.exports = {
     'no-console': 'warn',
     'no-plusplus': 'off',
     'no-restricted-syntax': 'off',
-    'no-unused-vars': 'warn',
     'no-var': 'error',
     'prettier/prettier': 'warn',
     'radix': 'off'
--- a/packages/mds-provider/api.ts
+++ b/packages/mds-provider/api.ts
@@ -84,7 +84,9 @@ function api(app: express.Express): express.Express {
         log.info(providerName(provider_id), req.method, req.originalUrl)
       }
     } catch (err) {
-      log.error(req.originalUrl, 'request validation fail:', err.stack)
+      const desc = err instanceof Error ? err.message : err
+      const stack = err instanceof Error ? err.stack : desc
+      log.error(req.originalUrl, 'request validation fail:', desc, stack || JSON.stringify(err))
     }
     next()
   })
@@ -139,9 +141,11 @@ function api(app: express.Express): express.Express {
         }
       )
     } catch (err) /* istanbul ignore next */ {
-      log.error('/test/seed failure:', err.stack)
+      const desc = err instanceof Error ? err.message : err
+      const stack = err instanceof Error ? err.stack : desc
+      log.error('/test/seed failure:', desc, stack || JSON.stringify(err))
       res.status(500).send({
-        result: `Failed to seed: ${err.stack}`
+        result: `Failed to seed: ${desc}`
       })
     }
   })
@@ -165,9 +169,11 @@ function api(app: express.Express): express.Express {
         }
       )
     } catch (err) /* istanbul ignore next */ {
-      log.error('/test/seed failure:', err.stack)
+      const desc = err instanceof Error ? err.message : err
+      const stack = err instanceof Error ? err.stack : desc
+      log.error('/test/seed failure:', desc, stack || JSON.stringify(err))
       res.status(500).send({
-        result: `Failed to seed: ${err.stack}`
+        result: `Failed to seed: ${desc}`
       })
     }
   })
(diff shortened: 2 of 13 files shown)


.eslintrc.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
     es6: true,
     mocha: true,
   },
-  plugins: ['@typescript-eslint', 'prettier'],
+  plugins: ["import", "prettier", "@typescript-eslint"],
```
avatarneil: What's the motivation for needing the `import` plugin?

packages/mds-provider/api.ts
```diff
@@ -199,6 +199,25 @@ function api(app: express.Express): express.Express {
 
   // / /////////////////////// trips /////////////////////////////////
 
+  const getStage0Properties = (items: { recorded: Timestamp; sequence?: number | null }[]) => {
```
avatarneil: I'm not a huge fan of numbers in function names, but considering we don't have very strictly defined style rules yet I guess it's okay!

packages/mds-provider/api.ts
```diff
@@ -352,49 +405,26 @@ function api(app: express.Express): express.Express {
...
-          version: PROVIDER_VERSION,
-          data: {
-            trips
-          },
-          links: asJsonApiLinks(req, skip, take, count)
-        })
+    } catch (err) {
+      res.status(500).send({
+        error: 'internal_failure',
+        error_description: `trips error: ${err.stack}`
```
avatarneil: You should never send the full error stack trace back to the client, that's potentially a massive security hole. The stack trace should only ever be output through `mds-logger`, like `log.error(err)`.

packages/mds-provider/api.ts
```diff
@@ -352,49 +405,26 @@ function api(app: express.Express): express.Express {
...
-              })
-            }, fail)
-            .catch(fail)
-        }, fail)
-        .catch(fail)
     }
+  }
+
+  app.get(pathsFor('/trips'), async (req: ProviderApiRequest, res: ProviderApiResponse) => {
+    await (req.query.newSkool ? getTrips(req, res) : getEventsAsTrips(req, res))
```
avatarneil: Probably a good time to rename this query param...

packages/mds-provider/event-processor/index.ts
```diff
@@ -16,69 +16,118 @@
...
 import { ProviderLabeler } from './labelers/provider-labeler'
 import { StreamEntry } from './types'
-import { StatusChangesProcessor, StatusChangesProcessorStreamEntry } from './processors/status-changes-processor'
+import { StatusChangesProcessor } from './processors/status-changes-processor'
 import { TripLabeler } from './labelers/trip-labeler'
-import { TripsProcessor, TripsProcessorStreamEntry, TripEvent } from './processors/trips-processor'
+import { TripsProcessor, TripEvent } from './processors/trips-processor'
 
-const isStatusChangesProcessorStreamEntry = (entry: StreamEntry): entry is StatusChangesProcessorStreamEntry =>
+const isStatusChangesProcessorStreamEntry = (entry: StreamEntry): entry is StreamEntry<VehicleEvent> =>
```
avatarneil: These functions should be broken out into a `validators.ts` file.

packages/mds-stream/index.ts
```diff
@@ -179,6 +207,38 @@ async function readStreamGroup(
...
+      groups,
+      ,
+      lastGeneratedId,
+      ,
+      firstEntry,
+      ,
+      lastEntry
+    ] = await client.xinfoAsync('STREAM', stream)
+    return { length, radixTreeKeys, radixTreeNodes, groups, lastGeneratedId, firstEntry, lastEntry }
+  } catch (err) {
```
avatarneil: We should probably at least log if an error occurs (unless an error occurring here isn't something to care about)

types/mds.d.ts
```diff
@@ -36,6 +36,8 @@ export interface Device {
 
 export type DeviceID = Pick<Device, 'provider_id' | 'device_id'>
 
+export type VehicleEventPrimaryKey = { timestamp: Timestamp; device_id: UUID } | null
```
avatarneil: I think this should be in types for `mds-db`

.eslintrc.js
mdurling: This plugin intends to support linting of ES2015+ (ES6+) import/export syntax, and prevent issues with misspelling of file paths and import names. All the goodness that the ES2015+ static module syntax intends to provide, marked up in your editor.

mdurling: Having said that, we could remove this plugin if we don't think it provides value.

packages/mds-provider/api.ts
mdurling: These are properties that are not in the MDS spec. Adopted "Stage 0" for proposals which is similar to what TC39 does for ECMAScript.

packages/mds-provider/api.ts
mdurling: I think I inherited this, but I will amend.

packages/mds-provider/api.ts
mdurling: It is going to go away completely. It exists to turn on/off the experimental provider API.

packages/mds-provider/event-processor/index.ts
mdurling: They are only used in this one module. They are useful outside of the event processor context.

packages/mds-stream/index.ts
mdurling: It's not. It will only throw if the stream doesn't exist, which is possible the first time this thing runs.

types/mds.d.ts
mdurling: I could move it, but it made sense to define it beside where `VehicleEvent` is declared.  Much of the stuff in `mds.d.ts` should move to `mds-db`, but @janedotx says there's a circular dependency between `mds-utils` and `mds-db` that needs to be resolved for that to happen. That should be a separate refactor to assess what's going on there.

packages/mds-provider/api.ts
mdurling: At some point we should move all of the various exception/error types into a common place and standardize the logging/handling. And perhaps add a function that can convert an `Error` into something safe to return as a response.

janedotx: Yeah, `mds-logger` censors latitude and longitude, which is potentially sensitive information.

PR #20, opened 2019-07-09
[mds-providers] Add constants, types, and type guards for provider_id

mdurling:
Updated provider information to latest version of CSV in [mobility-data-specification](https://github.com/CityOfLosAngeles/mobility-data-specification/blob/dev/providers.csv).

Also, added some additional constants, types, and type guards for working with provider_ids.


packages/mds-test-data/index.ts
```diff
@@ -303,7 +305,7 @@ function makeStatusChange(device: Device, timestamp: Timestamp, provider_name =
   }
 }
 
-function makeTrip(device: Device, provider_id = TEST_UUID, provider_name = 'test_provider'): Trip {
+function makeTrip(device: Device, provider_id = TEST1_PROVIDER_ID, provider_name = 'test_provider'): Trip {
```
avatarneil: I think `provider_name` here could be a call to `providerName`, something like `provider_name = providerName(TEST1_PROVIDER_ID)`

mdurling: Fixed.

PR #21, opened 2019-07-09
[mds-compliance] Filter out events older than 48hrs in a compliance context.

avatarneil:
LADOT would like to filter out events which are older than 48hrs at the time of compliance running, (i.e. if a device hasn't sent any events within 48 hours, then don't consider the most recent event). In the case that an `end_date` is specified in a call to `/snapshot`, we will consider a 48 hour window ending at the provided `end_date`.

*THIS IS TEMPORARY, AND NOT INTENDED TO END UP IN THE SPEC*

## PR Checklist

## Impacts
- [ ] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [x] Compliance


packages/mds-compliance/mds-compliance-engine.ts
```diff
@@ -276,4 +274,10 @@ function filterPolicies(policies: Policy[]): Policy[] {
   })
 }
 
-export { processPolicy, filterPolicies }
+function filterEvents(events: VehicleEvent[], end_time = now()): VehicleEvent[] {
+  return events.filter((event: VehicleEvent) => {
+    return event.timestamp > end_time - TWO_DAYS_IN_MS
```
karcass: I think your > should be <

karcass: recommend a comment for the intent, e.g. `keep events that are less than two days old`

PR #23, opened 2019-07-10
[mds-provider] event processing

mdurling:
Latest enhancements and bug fixes.

## Impacts
- [x] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance


packages/mds-db/index.ts
```diff
@@ -1513,16 +1522,19 @@ async function readEventsRangeExclusive(
...
   )
   return rows.map(({ lat, lng, telemetry_timestamp, ...event }) => ({
     ...event,
     telemetry_timestamp,
-    telemetry: {
-      timestamp: telemetry_timestamp,
-      gps: { lat, lng }
-    }
+    telemetry:
+      lat !== null && lng !== null
```
avatarneil: Could just be: `telemetry: lat && lng ? { timestamp: telemetry_timestamp, gps: { lat, lng } } : null`

packages/mds-provider/api.ts
```diff
@@ -221,66 +204,56 @@ function api(app: express.Express): express.Express {
...
-      properties: {
-        timestamp: item.timestamp
-      },
-      geometry: {
-        type: 'Point',
-        coordinates: [round(item.gps.lng, 6), round(item.gps.lat, 6)]
-      }
-    }
-  }
+  // function asFeature(item: Telemetry): Feature {
```
avatarneil: Why commenting out versus nuking/moving to `mds-utils`?

packages/mds-provider/api.ts
```diff
@@ -221,66 +204,56 @@ function api(app: express.Express): express.Express {
     return undefined
```
avatarneil: We don't need to return anything here I think

packages/mds-provider/event-processor/index.ts
```diff
@@ -123,28 +117,53 @@ async function process(options: ReadStreamOptions): Promise<void> {
...
+const stop = async (): Promise<void> => {
   logger.info('Stopping Event Processor')
   await Promise.all([stream.shutdown(), db.shutdown()])
 }
 
-export type ProviderEventProcessorOptions = ReadStreamOptions
+const wait = async (interval: number): Promise<number> =>
+  interval > 0 ? new Promise(resolve => setTimeout(() => resolve(interval), interval)) : Promise.resolve(interval)
+
+/* eslint-disable no-await-in-loop */
```
avatarneil: Add an `/* eslint-reason ...*/` please

packages/mds-provider/package.json
```diff
@@ -14,8 +14,8 @@
     "test": "yarn test:prettier && yarn test:eslint && yarn test:unit",
     "test:eslint": "eslint --ignore-path ../../.gitignore '**/*.ts'",
     "test:prettier": "prettier --write --check --ignore-path ../../.gitignore '**/*.ts'",
-    "test:unit": "nyc --check-coverage --exclude tests --extension .ts --lines 80 --reporter=text --reporter=html ts-mocha --require source-map-support/register --recursive tests/**/*.ts",
-    "watch": "nodemon --watch '../../packages' --ext 'ts' --ignore '*.d.ts' --exec yarn watch:exec",
+    "test:unit": "nyc --check-coverage --exclude tests --extension .ts --lines 75 --reporter=text --reporter=html ts-mocha --require source-map-support/register --recursive tests/**/*.ts",
```
avatarneil: I don't really like the idea of reducing test coverage % significantly with the addition of new features... We should either write new tests which have more coverage, or add istanbul annotations if the code can't be reasonably tested.

packages/mds-db/index.ts
mdurling: I guess that's safe ... as long as you want to treat `{ lat: 0, lng: 0 }` as `null` also.

packages/mds-provider/api.ts
mdurling: They will be making a comeback.

packages/mds-provider/api.ts
mdurling: Actually, I think `return {}` is probably more appropriate.

packages/mds-provider/package.json
mdurling: I did add some new tests and increased the % ... I will update this to `--lines 85`

packages/mds-db/index.ts
karcass: or undefined :)

mdurling: Coming from db it will never be undefined

karcass: doh

packages/mds-db/index.ts
```diff
@@ -1266,6 +1266,14 @@ async function readStatusChanges(
     }
   }
 
+  if (device_id) {
+    if (!isUUID(device_id)) {
```
janedotx: Real dumb question, but isn't the type declaration supposed to ensure this is a UUID?

mdurling: If I understand your question correctly, ... The `UUID` type is just an alias for `string`. Regardless though, static type checking only helps at compilation time. At run time, you have to type guard everything. That's why most of the validators I've written take an input argument of type `arg: unknown` and returns `arg is <type>`. Which is basically, accept this argument of unknown type, apply some tests/constraints to it, and it they all pass, treat it as a `<type>` from that point forward in the code.

janedotx: Ah right, that makes sense. Thanks.

PR #32, opened 2019-07-11
[mds-provider] remove sequence column

mdurling:
By guaranteeing query order (recorded IDX + PK), we can use the existing `skip/take` parameters to reliably poll for new status changes without the need for an additional sequencing mechanism.

## Impacts
- [x] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance

### DIFF
--- a/packages/mds-db/index.ts
+++ b/packages/mds-db/index.ts
@@ -1472,41 +1472,23 @@ async function readRule(rule_id: UUID): Promise<Rule> {
   }
 }
 
-async function getMostRecentStatusChange(): Promise<Recorded<StatusChange> | null> {
-  // SELECT * FROM status_changes ORDER BY event_time DESC, device_id DESC LIMIT 1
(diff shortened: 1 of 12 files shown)


packages/mds-db/index.ts
```diff
@@ -1178,7 +1170,7 @@ async function readTrips(
   }
 
   const { rows: trips } = await exec(
-    `SELECT * FROM ${schema.TRIPS_TABLE} ${where} ORDER BY recorded ASC, sequence ASC${
+    `SELECT * FROM ${schema.TRIPS_TABLE} ${where} ORDER BY recorded ASC, provider_trip_id ASC${
```
avatarneil: `provider_trip_id` seems like a kind of odd thing to be sorting by... Necessary?

mdurling: Sorts are `recorded IDX + PK`

PR #33, opened 2019-07-11
[mds-api-server] Remove redundancy using standard authorizer claims

mdurling:
Each API implemented its own authorization header parser. Converted APIs to utilize the standard authorizer claims provided by `mds-api-authorizer` and `mds-api-server and removed of all the various and largely redundant `getAuth` functions.

## Impacts
- [x] Provider
- [x] Agency
- [x] Audit
- [x] Policy
- [x] Compliance

### DIFF
--- a/packages/mds-api-authorizer/tests/index.spec.ts
+++ b/packages/mds-api-authorizer/tests/index.spec.ts
@@ -9,13 +9,13 @@ const ADMIN_AUTH = `basic ${Buffer.from(`${PROVIDER_UUID}|${PROVIDER_SCOPES}`).t
 describe('Test API Authorizer', () => {
   it('No Authorizaton', done => {
     const authorizer = AuthorizationHeaderApiAuthorizer({} as express.Request)
-    test.value(authorizer).is({})
+    test.value(authorizer).is(null)
     done()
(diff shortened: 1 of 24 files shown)


packages/mds-api-authorizer/tests/index.spec.ts
```diff
@@ -9,13 +9,13 @@ const ADMIN_AUTH = `basic ${Buffer.from(`${PROVIDER_UUID}|${PROVIDER_SCOPES}`).t
 describe('Test API Authorizer', () => {
   it('No Authorizaton', done => {
     const authorizer = AuthorizationHeaderApiAuthorizer({} as express.Request)
-    test.value(authorizer).is({})
+    test.value(authorizer).is(null)
```
janedotx: Just curious, what's the motivation for this change?

mdurling: Makes the type guarding more predictable because it forces authorizers to return claims and not partial claims and is more reflective of what actually happens in the authorizer.

janedotx: 👍

PR #34, opened 2019-07-11
[mds-agency] Optimize POST event query

avatarneil:
Instead of reading all device_ids to verify proper provider when POSTing an event, read the device and check for matching provider_id.

## PR Checklist

## Impacts
- [ ] Provider
- [x] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance

### DIFF
--- a/packages/mds-agency/api.ts
+++ b/packages/mds-agency/api.ts
@@ -832,6 +832,7 @@ function api(app: express.Express): express.Express {
     const { device_id } = req.params
 
     const { provider_id } = getAuth(req)
+    const name = providerName(provider_id || 'unknown')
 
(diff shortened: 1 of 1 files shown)


packages/mds-agency/api.ts
```diff
@@ -832,6 +832,7 @@ function api(app: express.Express): express.Express {
     const { device_id } = req.params
 
     const { provider_id } = getAuth(req)
+    const name = providerName(provider_id || 'unknown')
```
mdurling: This does't look right. `const name = providerName(provider_id) || 'unknown'` perhaps?

PR #36, opened 2019-07-11
[mds-policy][mds-db] DRAFT: Implement policy publishing and editing

janedotx:
## PR Checklist

## Impacts
- [ ] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance


janedotx:
@avatarneil I should probably do the admin endpoints for publishing too, eh? I just realized that as I cut this PR. But I'd appreciate a sanity check from you before deeper into this.

@karcass What's with that bit about `unpublished` having no effect if you have a `provider_id` in your token? If you don't have a `provider_id`, these endpoints won't authorize you at the moment. Should that change?

packages/mds-db/tests/mds-db.spec.ts
```diff
@@ -228,5 +230,50 @@ if (pg_info.database) {
...
+        policy.name = 'a shiny new name'
+        await MDSDBPostgres.editPolicy(policy)
+        const result = await MDSDBPostgres.readPolicies({ policy_id: POLICY2_JSON.policy_id })
+        assert.deepEqual(result[0].name, 'a shiny new name')
+      })
+
+      it('will not edit a published Policy', async () => {
+        const publishedPolicy = clone(POLICY_JSON)
+        publishedPolicy.name = 'a shiny new name'
+        await MDSDBPostgres.editPolicy(publishedPolicy)
```
avatarneil: `editPolicy` should probably throw an error if you try to edit a published policy, I think it'd be better to verify that an error is thrown instead of checking that the string for name hasn't changed.

packages/mds-policy/api.ts
```diff
@@ -133,15 +133,19 @@ function api(app: express.Express): express.Express {
...
+    const { start_date = now(), end_date = now(), unpublished } = req.query
     log.info('read /policies', req.query, start_date, end_date)
     if (start_date > end_date) {
       res.status(400).send({ result: 'start_date after end_date' })
       return
     }
-    db.readPolicies({ start_date, end_date })
+
+    let get_unpublished = false
+    if (unpublished !== undefined || unpublished !== null) {
```
avatarneil: I don't think `unpublished` should be something that's part of the query... `unpublished` policies should be viewable to a specific audience (i.e. policy authors). I think this makes more sense to have as an entirely separate endpoint, as this change would differ from the current Policy spec for the `/policies` endpoint.

packages/mds-policy/api.ts
```diff
@@ -243,6 +247,18 @@ function api(app: express.Express): express.Express {
     })
     .unknown(true) // TODO
 
+  app.post(pathsFor('/policies/:policy_id'), (req, res) => {
```
avatarneil: Stick this behind an admin endpoint for the time being, don't want this publicly exposed.

packages/mds-policy/api.ts
```diff
@@ -361,6 +377,31 @@ function api(app: express.Express): express.Express {
       })
   })
 
+  app.post(pathsFor('/admin/policies/:policy_id/publish'), (req, res) => {
```
avatarneil: I may be misunderstanding the behavior of this endpoint, but it looks like you're writing an unpublished policy to the db, not publishing it. Needs a rename

avatarneil (commented):
Looks like a good start

packages/mds-policy/api.ts
avatarneil: And it's important that unpublished policies are only viewable by the author as an audience.

janedotx: I mean, it's in the draft API spec. "The only addition is that when calling the GET /policies endpoint, append ?unpublished to get the unpublished ones, separate from the published ones.  Adding ?unpublished will have no effect if your access token has a provider_id.
"
@karcass what's the motivation for this query param?

packages/mds-policy/api.ts
janedotx: 👍

packages/mds-policy/api.ts
janedotx: You are so right, I derped there.

packages/mds-db/tests/mds-db.spec.ts
janedotx: done

PR #39, opened 2019-07-12
Standardize webpack configuration to support multiple bundles

mdurling:
There is now a simpler and more standardized way to build deployable bundles with less replicated configuration. The change also supports building more than one bundle per package. Also, the deployment scripts have been updated to match bundles and regions to specific AWS Lambda functions.


aws-lambda/mds-metrics-sheet/package.json
```diff
@@ -7,8 +7,10 @@
   "scripts": {
     "build": "tsc --build .",
     "bundle": "webpack --mode=production --env.npm_package_name=${npm_package_name} --env.npm_package_version=${npm_package_version}",
-    "deploy:us-east-2": "aws lambda update-function-code --region us-east-2 --zip-file fileb://dist/${npm_package_name}.zip --function-name",
-    "deploy:us-west-2": "aws lambda update-function-code --region us-west-2 --zip-file fileb://dist/${npm_package_name}.zip --function-name",
+    "deploy:dev3-metrics-sheet": "yarn lambda:us-east-2 dev3-metrics-sheet --zip-file fileb://dist/bundles/metrics-log.zip",
```
avatarneil: 👏

PR #40, opened 2019-07-12
[mds-compliance] Remove error throw

avatarneil:
Don't throw an error when a device is in DB and not in cache when computing realtime compliance snapshots.

## PR Checklist

## Impacts
- [ ] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [x] Compliance

### DIFF
--- a/packages/mds-compliance/api.ts
+++ b/packages/mds-compliance/api.ts
@@ -176,10 +176,6 @@ function api(app: express.Express): express.Express {
                           /* istanbul ignore next */
                           const deviceMap: { [d: string]: Device } = devices.reduce(
                             (deviceMapAcc: { [d: string]: Device }, device: Device) => {
-                              /* istanbul ignore next */
-                              if (!device) {
-                                return deviceMapAcc
-                              }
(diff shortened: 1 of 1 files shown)


packages/mds-compliance/api.ts
```diff
@@ -178,7 +178,7 @@ function api(app: express.Express): express.Express {
                             (deviceMapAcc: { [d: string]: Device }, device: Device) => {
                               /* istanbul ignore next */
                               if (!device) {
-                                throw new Error('device in DB but not in cache')
+                                return deviceMapAcc
                               }
                               return Object.assign(deviceMapAcc, { [device.device_id]: device })
```
mdurling: If you're not going to throw, you don't even need the `if` ... `Object.assign` will just put that device_id in the map as undefined.

avatarneil: Ah, cool beans! I'll get rid of that `if` then.

mdurling: Alternately, just `return device ? Object.assign(deviceMapAcc, { [device.device_id]: device }) : deviceMapAcc`

PR #44, opened 2019-07-15
[mds-agency][mds-provider] add register event

karcass:
Northbound Provider and other feeds should include a `register` event, per spec, but we will add that as part of the registration rather than asking providers to send a separate event

### DIFF
--- a/packages/mds-agency/tests/test.ts
+++ b/packages/mds-agency/tests/test.ts
@@ -89,7 +89,7 @@ let testTimestamp = now()
 
 const test_event = {
   device_id: DEVICE_UUID,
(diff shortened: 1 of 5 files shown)


karcass:
Probably should add a test for this :)

packages/mds-agency/tests/test.ts
```diff
@@ -89,7 +89,7 @@ let testTimestamp = now()
 
 const test_event = {
   device_id: DEVICE_UUID,
-  event_type: 'deregister',
+  event_type: VEHICLE_EVENTS.deregister,
```
janedotx: 👍

PR #45, opened 2019-07-15
[mds-*] Add promise linter rules, and clean up half of agency

avatarneil:
## PR Checklist

## Impacts
- [x] Provider
- [x] Agency
- [x] Audit
- [x] Policy
- [x] Compliance

### DIFF
--- a/packages/mds-agency/api.ts
+++ b/packages/mds-agency/api.ts
@@ -164,18 +164,19 @@ function api(app: express.Express): express.Express {
    * See {@link https://github.com/CityOfLosAngeles/mobility-data-specification/tree/dev/agency#service_areas Service Areas}
    */
   app.get(pathsFor('/service_areas'), async (req: AgencyApiRequest, res: AgencyApiResponse) => {
-    const serviceAreas = await areas.readServiceAreas().catch((err: Error) => {
+    try {
+      const serviceAreas = await areas.readServiceAreas()
+      log.info('readServiceAreas (all)', serviceAreas.length)
+      return res.status(200).send({
+        service_areas: serviceAreas
+      })
+    } catch (err) {
       /* istanbul ignore next */
       log.error('failed to read service areas', err)
       return res.status(404).send({
         result: 'not found'
       })
-    })
-
-    log.info('readServiceAreas (all)', serviceAreas.length)
-    return res.status(200).send({
-      service_areas: serviceAreas
-    })
+    }
   })
 
   /**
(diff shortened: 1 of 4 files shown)


packages/mds-agency/api.ts
```diff
@@ -163,28 +163,26 @@ function api(app: express.Express): express.Express {
...
-      },
-      (err: object) => /* istanbul ignore next */ {
-        log.error('failed to read service areas', err)
-        res.status(404).send({
-          result: 'not found'
-        })
-      }
-    )
+  app.get(pathsFor('/service_areas'), async (req: AgencyApiRequest, res: AgencyApiResponse) => {
+    const serviceAreas = await areas.readServiceAreas().catch((err: Error) => {
```
mdurling: I don't like mixing await with promises in this manner. In this case `serviceAreas` will probably be inferred as `ServiceArea[] | void` which makes typing of responses problematic. IMO, a better pattern is:
```
try {
  const serviceAreas = await areas.readServiceAreas()
  log.info('readServiceAreas (all)', serviceAreas.length)
  return res.status(200).send({ service_areas: serviceAreas })
} 
catch (err: Error) {
  log.error('failed to read service areas', err)
  return res.status(404).send({
    result: 'not found'
  })
}
```

avatarneil: @mdurling What do you think about the use of `.catch` with `await` in `/vehicles/:device_id`?

mdurling: I only pointed out one instance of the pattern. The others should be changed as well. Mixing `await` with `.catch` will lead to type confusion at some level.

avatarneil: The original instance of the pattern you pointed out made sense to resolve, however I don't see the reasoning with not using `.catch` after an `await`. There are some cases where you want to handle a rejection from promise `a` differently than promise `b`, as is demonstrated quite well by the `/vehicles/:device_id` example. Considering we don't have any sort of error code specification, and a function may only reject and not throw an error, I think handling it like this is the best we can do. Any type confusion that would arise can always be resolved by type-guarding, ```typescript
const res = await someAsyncFunction().catch(err => {
  log.warn(err)
  return undefined
})
if (!(res instanceOf ExpectedType)) {
  ...
}
rest of code that depends on res
```

If you can find a clean, concise way to handle it using just `try/catch` blocks with `async/await`, I'd love to see it.

mdurling:
A function that rejects should still be caught in try/await/catch. In my experience await/.catch has always been replaceable with try/await/catch, but I defer to your judgement on this particular case.

PR #49, opened 2019-07-15
[mds-db] [mds-agency] Optimize writeTelemetry

avatarneil:
In the case that a provider hits the `/vehicles/telemetry/` endpoint and only provides one telemetry datapoint instead of a batch, don't read all device_ids for the provider.

## PR Checklist

## Impacts
- [ ] Provider
- [x] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance


packages/mds-db/index.ts
```diff
@@ -290,12 +290,16 @@ async function readDeviceIds(provider_id?: UUID, skip?: number, take?: number):
 }
 
 // TODO: FIX updateDevice/readDevice circular reference
-async function readDevice(device_id: UUID): Promise<Recorded<Device>> {
+async function readDevice(device_id: UUID, provider_id?: UUID): Promise<Recorded<Device>> {
   return new Promise((resolve, reject) => {
     // read from pg
     getReadOnlyClient().then(client => {
-      const sql = `SELECT * FROM ${schema.DEVICES_TABLE} WHERE device_id=$1`
+      let sql = `SELECT * FROM ${schema.DEVICES_TABLE} WHERE device_id=$1`
```
mdurling: Use `SqlVals` instead of hardcoding positional arguments `$1`, etc.

packages/mds-db/index.ts
```diff
@@ -290,12 +290,16 @@ async function readDeviceIds(provider_id?: UUID, skip?: number, take?: number):
...
       const values = [device_id]
+      if (provider_id) {
+        sql += ` AND provider_id=$2`
```
mdurling: This could be done as follows: ``const sql = `SELECT * FROM ${schema.DEVICES_TABLE} WHERE device_id=${vals.add(device_id)}${provider_id ? ` AND provider_id=${vals.add(provider_id)}` : ''}``

avatarneil: That's what I said! Max replies: 'Sure, and we could do everything in one line of APL too!'

mdurling: @karcass Knows APL? I took a whole semester of it :-)

mdurling: It's not so much that I dislike multiple lines; I just dislike `let`.

packages/mds-agency/api.ts
```diff
@@ -533,7 +533,7 @@ function api(app: express.Express): express.Express {
     const { provider_id } = res.locals
 
     function fail(err: Error | string): void {
-      log.error('readDeviceIds fail', err).then(() => {
+      log.error('readDeviceIds fail', JSON.stringify(err)).then(() => {
```
mdurling: Should just log `err` ... we should go through and remove most of the `JSON.stringify` calls in logger parameters.

packages/mds-agency/api.ts
```diff
@@ -958,16 +952,22 @@ function api(app: express.Express): express.Express {
...
       })
       return
     }
     const failures: string[] = []
     const valid: Telemetry[] = []
 
     const recorded = now()
-
-    db.readDeviceIds(provider_id).then((device_ids: DeviceID[]) => {
+    let p: Promise<Device | DeviceID[]>
```
mdurling: Make this `const` and use a ternary expression.

PR #50, opened 2019-07-16
[mds-*] never log '[object Object]'

karcass:
convert using JSON.stringify

### DIFF
--- a/packages/mds-logger/index.ts
+++ b/packages/mds-logger/index.ts
@@ -115,7 +115,11 @@ function makeCensoredLogMsgRecurse(msg: { [propName: string]: any }): { [propNam
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 function makeCensoredLogMsg(...msg: any) {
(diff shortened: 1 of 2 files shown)


packages/mds-logger/index.ts
```diff
@@ -115,7 +115,11 @@ function makeCensoredLogMsgRecurse(msg: { [propName: string]: any }): { [propNam
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 function makeCensoredLogMsg(...msg: any) {
-  return makeCensoredLogMsgRecurse(msg)
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  return makeCensoredLogMsgRecurse(msg).map((item: any) =>
+    // never print out '[object Object]'
+    String(item) === '[object Object]' ? JSON.stringify(item) : item
```
mdurling: Would `typeof item === 'object'` have worked here as well?

PR #51, opened 2019-07-16
[mds-native] [mds-audit] Native event feed

mdurling:
1. Implement new API/endpoint `GET /native/events` for reading agency events.
2. Moved validators/exceptions into `[mds-utils]` for better reusability.
3. Reorganized `[mds-audit]` to match other packages since it was an outlier.
4. Upgraded dependencies and refined some typings as required by new TypeScript version.
5. Implement new API/endpoint `GET /native/devices/:device_id` for reading device registration.

## Impacts
- [ ] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance
- [x] Native


packages/mds-native/api.ts
```diff
@@ -0,0 +1,101 @@
...
+import express from 'express'
+import { pathsFor, isValidProviderId, ValidationError, ServerError, AuthorizationError } from 'mds-utils'
+import logger from 'mds-logger'
+import db from 'mds-db'
+import { NextFunction } from 'connect'
+import { providerName } from 'mds-providers'
+import { asPagingParams, asJsonApiLinks } from 'mds-api-helpers'
+import { NativeApiResponse, NativeApiRequest, NativeApiGetEventsRequest, NativeApiGetEventsReponse } from './types'
+
+const NATIVE_API_VERSION = '0.0.1'
```
avatarneil: We should be able to get this from `package.json`

mdurling: It's not the same thing. This is the API format version, not the code version.

packages/mds-db/index.ts
```diff
@@ -384,7 +384,7 @@ async function updateDevice(device_id: UUID, changes: Partial<Device>): Promise<
 
 async function writeEvent(event_param: VehicleEvent) {
   await readDevice(event_param.device_id)
-  const client = await getWriteableClient() 
+  const client = await getWriteableClient()
```
avatarneil: Oopsie, prettier didn't want to work :P

PR #52, opened 2019-07-16
[mds-agency] Improve error handling when POSTing a vehicle

avatarneil:
## PR Checklist

## Impacts
- [ ] Provider
- [x] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance

### DIFF
--- a/packages/mds-agency/api.ts
+++ b/packages/mds-agency/api.ts
@@ -307,7 +307,7 @@ function api(app: express.Express): express.Express {
       mfgr: body.mfgr,
       model: body.model,
       recorded,
-      status: undefined
+      status: VEHICLE_STATUSES.removed
     }
 
     const failure = badDevice(device)
@@ -347,7 +347,6 @@ function api(app: express.Express): express.Express {
     try {
       await db.writeDevice(device)
       await Promise.all([cache.writeDevice(device), stream.writeDevice(device)])
-      device.status = VEHICLE_STATUSES.removed
       await log.info('new', providerName(res.locals.provider_id), 'vehicle added', JSON.stringify(device))
       await writeRegisterEvent()
       res.status(201).send({ result: 'register device success', recorded, device })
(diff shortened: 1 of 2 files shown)


packages/mds-agency/api.ts
```diff
@@ -316,73 +316,51 @@ function api(app: express.Express): express.Express {
...
+      await writeRegisterEvent()
+      res.status(201).send({ result: 'register device success', recorded, device })
+    } catch (err) {
+      if (String(err).includes('duplicate')) {
+        res.status(409).send({
+          error: 'already_registered',
+          error_description: 'A vehicle with this device_id is already registered'
+        })
+      } else if (String(err).includes('db')) {
+        res.status(500).send(new ServerError())
```
karcass: why no log in this block but log in next block?

packages/mds-agency/api.ts
```diff
@@ -316,73 +316,51 @@ function api(app: express.Express): express.Express {
...
-          })
-        } else {
-          log.error(providerName(res.locals.provider_id), 'register vehicle failed:', err).then(() => {
-            res.status(500).send(new ServerError())
-          })
-        }
+    try {
+      await db.writeDevice(device)
+      await Promise.all([cache.writeDevice(device), stream.writeDevice(device)])
+      device.status = VEHICLE_STATUSES.removed
```
mdurling: The device has already been written to db/cache/stream .. what does setting the status here accomplish? In general, I would prefer `res.status(201).send({ result: 'register device success', recorded, device: { ...device, status: VEHICLE_STATUSES.removed })`

packages/mds-agency/api.ts
```diff
@@ -316,73 +316,51 @@ function api(app: express.Express): express.Express {
...
+        telemetry: null,
+        timestamp: recorded,
+        trip_id: null,
+        recorded,
+        telemetry_timestamp: undefined,
+        service_area_id: null
+      }
+      try {
+        await db.writeEvent(event)
+        await Promise.all([cache.writeEvent(event), stream.writeEvent(event)]).catch(async err => {
```
mdurling: This could be within a `try/catch` instead of using `Promise.catch`, right?

karcass: done

packages/mds-agency/api.ts
karcass: probably that mutation should happen at the top

PR #53, opened 2019-07-16
[mds-db] Refactor writeEvent to async/await

avatarneil:
## PR Checklist

## Impacts
- [ ] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance

### DIFF
--- a/packages/mds-db/index.ts
+++ b/packages/mds-db/index.ts
@@ -385,18 +385,14 @@ async function updateDevice(device_id: UUID, changes: Partial<Device>): Promise<
 
 async function writeEvent(event_param: VehicleEvent) {
   const device = await readDevice(event_param.device_id)
-  if (!device) {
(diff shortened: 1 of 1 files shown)


packages/mds-db/index.ts
```diff
@@ -383,28 +383,20 @@ async function updateDevice(device_id: UUID, changes: Partial<Device>): Promise<
...
-        client
-          .query(sql, values)
-          .then(() => {
-            resolve(event as Recorded<VehicleEvent>)
-          }, reject)
-          .catch(reject)
-      })
-    }
-  })
+  if (!device) {
```
mdurling: I am pretty sure `readDevice` will `reject` so this `!device` may never happen. You should test that.

avatarneil: You're correct! I think this is an artifact from long ago.

PR #57, opened 2019-07-17
[mds-*] support build on install

twelch:
In order to use mds-core as a library, the importing module needs access to mds-core build artifacts under `dist` for each package.  But those build artifacts are not published anywhere at this time.

Short of publishing mds-core to a package registry, this PR implements the solution of using a `prepare` script to tell the importing library to build mds-core on install and put the build artifacts into node_modules/mds-core alongside the src.

A side effect of `prepare` is that it also gets run after every `yarn install` within mds-core.

In addition, a `files` include is used to workaround a yarn bug in which the prepare lifecycle script is not run for a git hosted dependency.
https://github.com/yarnpkg/yarn/issues/5235#issuecomment-490019287

## PR Checklist

## Impacts
- [ ] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   ],
   "files": [
     "*",
-    "packages/*/dist"
+    "packages/*/dist/"
   ],
   "devDependencies": {
     "@types/bluebird": "3.5.27",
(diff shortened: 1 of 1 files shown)


avatarneil:
Can one of the admins verify this patch?

PR #59, opened 2019-07-18
[mds-db] Add useWriteable param to readDevice to mitigate race condition for device registration

avatarneil:
## PR Checklist

## Impacts
- [ ] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance

### DIFF
--- a/packages/mds-db/index.ts
+++ b/packages/mds-db/index.ts
@@ -246,8 +246,8 @@ async function readDeviceIds(provider_id?: UUID, skip?: number, take?: number): 
 }
 
 // TODO: FIX updateDevice/readDevice circular reference
-async function readDevice(device_id: UUID, useWriteable?: boolean): Promise<Recorded<Device>> {
-  const client = useWriteable ? await getWriteableClient() : await getReadOnlyClient()
+async function readDevice(device_id: UUID, optionalClient?: MDSPostgresClient): Promise<Recorded<Device>> {
+  const client = optionalClient || (await getReadOnlyClient())
   const sql = `SELECT * FROM ${schema.DEVICES_TABLE} WHERE device_id=$1`
   const values = [device_id]
   logSql(sql, values)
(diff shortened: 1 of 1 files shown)


packages/mds-db/index.ts
```diff
@@ -246,8 +246,8 @@ async function readDeviceIds(provider_id?: UUID, skip?: number, take?: number):
 }
 
 // TODO: FIX updateDevice/readDevice circular reference
-async function readDevice(device_id: UUID): Promise<Recorded<Device>> {
-  const client = await getReadOnlyClient()
+async function readDevice(device_id: UUID, useWriteable?: boolean): Promise<Recorded<Device>> {
```
mdurling: This feels like a bit of boolean plinko to me. It's not obvious when/why someone would want to use true. I have a suggestion. Instead of a boolean, why not use an optional `client` parameter instead; at least then when looking at it one might say "Oh, I already have a client; let me just reuse that". So below, then would be:
```
const client = await getWritableClient()
await readDevice(event_param.device_id, client)
```

packages/mds-db/index.ts
```diff
@@ -297,7 +297,7 @@ async function updateDevice(device_id: UUID, changes: Partial<Device>): Promise<
 }
 
 async function writeEvent(event_param: VehicleEvent) {
-  await readDevice(event_param.device_id)
+  await readDevice(event_param.device_id, true)
```
mdurling: So, is this just to make sure the device exists such that it will throw an exception if if doesn't? We could use FKs to enforce this sort of referential integrity.

packages/mds-db/index.ts
janedotx: It doesn't seem like the business of the DB layer's clients to know about writeable vs readable clients. `getWriteableClient` and `getReadOnlyClient` aren't exported right now for that reason.

mdurling: That extra parameter is only currently used within db, but you're right. Whether the parameter is a boolean or a client instance, it exposes control over the client to the callers either way and that's probably not ideal. I think @janedotx makes a good point and perhaps what we really need is a public `readDevice` method and an internal implementation. `readDevice` and `writeEvent` could both call into the private method and pass the client.

packages/mds-db/index.ts
janedotx: We should probably be using the cache more but it's probably best to wait till after we move to Kubernetes.

PR #60, opened 2019-07-18
[mds-agency] detect all-redundant telemetry

karcass:
return a 400 and add'l error info if no new telemetry was sent

### DIFF
--- a/packages/mds-agency/api.ts
+++ b/packages/mds-agency/api.ts
@@ -729,21 +729,20 @@ function api(app: express.Express): express.Express {
 
   async function writeTelemetry(telemetry: Telemetry | Telemetry[]) {
     if (!Array.isArray(telemetry)) {
-      const promises: (Promise<any>)[] = [
+      const promises: [Promise<number>, Promise<void>, Promise<void>] = [
         db.writeTelemetry([telemetry]),
         cache.writeTelemetry([telemetry]),
         stream.writeTelemetry([telemetry])
       ]
       return Promise.all(promises)
     }
-    const promises: Promise<any>[] = [
+    const promises: [Promise<number>, Promise<void>, Promise<void>] = [
       db.writeTelemetry(telemetry),
       cache.writeTelemetry(telemetry),
       stream.writeTelemetry(telemetry)
     ]
     return Promise.all(promises)
   }
-
   /**
    * Endpoint to submit vehicle events
    * See {@link https://github.com/CityOfLosAngeles/mobility-data-specification/tree/dev/agency#vehicle---event Events}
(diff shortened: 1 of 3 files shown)


packages/mds-agency/api.ts
```diff
@@ -113,7 +113,7 @@ function api(app: express.Express): express.Express {
           // stash provider_id
           res.locals.provider_id = provider_id
 
-          log.info(providerName(provider_id), req.method, req.originalUrl)
+          // log.info(providerName(provider_id), req.method, req.originalUrl)
```
avatarneil: Should just delete

packages/mds-agency/api.ts
```diff
@@ -727,16 +727,20 @@ function api(app: express.Express): express.Express {
     return null
   }
 
-  async function writeTelemetry(telemetry: Telemetry | Telemetry[]): Promise<void[]> {
+  async function writeTelemetry(telemetry: Telemetry | Telemetry[]) {
     if (!Array.isArray(telemetry)) {
-      const promises = [
+      const promises: (Promise<any>)[] = [
```
avatarneil: I'm gonna clone this down and see if I can find a better way to express the type than this hack we came up with earlier today... Seems super odd that we'd need to use the `Promise<any>[]` here. WE WILL REBUILD IT! MAKE IT STRONGER! FASTER!

mdurling: Try to use destructuring and let TS infer the types `const [x, y, z] = [...]`or declare the type as a tuple `const promises: [Promise<x>, Promise<y>, Promise<z>] = ...`

avatarneil:
```
async function writeTelemetry(telemetry: Telemetry | Telemetry[]) {
    if (!Array.isArray(telemetry)) {
      const promises: [Promise<number>, Promise<void>, Promise<void>] = [
        db.writeTelemetry([telemetry]),
        cache.writeTelemetry([telemetry]),
        stream.writeTelemetry([telemetry])
      ]
      return Promise.all(promises)
    }
    const promises: [Promise<number>, Promise<void>, Promise<void>] = [
      db.writeTelemetry(telemetry),
      cache.writeTelemetry(telemetry),
      stream.writeTelemetry(telemetry)
    ]
    return Promise.all(promises)
  }
```

seems to work!

packages/mds-agency/api.ts
karcass: well, we had it before, we might want it again later
generally though I agree

packages/mds-agency/api.ts
karcass: check it in!

PR #66, opened 2019-07-22
[mds-*] reduce kms file handle consumption

karcass:
Move KMS object creation out of the decrypt() call, so it'll only be one per lambda

Change log.{info|warn|error} to use more sensible signature of `any[]`

Switch bad-event from `warn` to `info` to reduce slack spam

### DIFF
--- a/packages/mds-cache/index.ts
+++ b/packages/mds-cache/index.ts
@@ -181,36 +181,35 @@ async function readDevicesStatus(query: { since?: number; skip?: number; take?: 
   const start = query.since || 0
   const stop = now()
   // read all device ids
-  log
-    .info('redis zrangebyscore device-ids', start, stop)(await getClient())
-    .zrangebyscoreAsync('device-ids', start, stop)
-    .then(async (device_ids_res: string[]) => {
-      log.info('readDevicesStatus', device_ids_res.length, 'entries')
+  log.info('redis zrangebyscore device-ids', start, stop)
+  const client = await getClient()
+  client.zrangebyscoreAsync('device-ids', start, stop).then(async (device_ids_res: string[]) => {
+    log.info('readDevicesStatus', device_ids_res.length, 'entries')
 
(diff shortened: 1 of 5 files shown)


packages/mds-logger/index.ts
```diff
@@ -186,29 +188,25 @@ function info(...msg: any) {
 }
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-async function warn(...msg: any) {
-  try {
-    if (env.QUIET) {
-      return
-    }
+async function warn(...msg: any[]): Promise<any[]> {
```
mdurling: Not required for this PR, as it probably would increase the footprint of this change, but I would be inclined to use `unknown` instead of `any` in these sorts of functions.
>“TypeScript 3.0 introduces a new type called unknown that does exactly that. Much like any, any value is assignable to unknown; however, unlike any, unknown is assignable to almost nothing else without a type assertion. You also can’t access any properties off of an unknown, nor can you call/construct them.”

packages/mds-cache/index.ts
```diff
@@ -181,36 +181,35 @@ async function readDevicesStatus(query: { since?: number; skip?: number; take?:
...
+  log.info('redis zrangebyscore device-ids', start, stop)
+  const client = await getClient()
+  client.zrangebyscoreAsync('device-ids', start, stop).then(async (device_ids_res: string[]) => {
+    log.info('readDevicesStatus', device_ids_res.length, 'entries')
+
+    const skip = query.skip || 0
+    const take = query.take || 100000000000
+    const device_ids = device_ids_res.slice(skip, skip + take)
+
+    // read all devices
```
mdurling: Not related to your changes, but there sure are a lot of mutable operations happening here.

karcass: Yep, needs a once-over for sure.

PR #69, opened 2019-07-22
[mds-*] Finished converting to async/await

avatarneil:
Finished converting everything over to async/await and made the relevant linter rules throw an error if violated. Also added eslint-disables for test files so they wouldn't complain about promise handling, as mocha relies on callbacks (it's a huge pain to use it without).

## PR Checklist

## Impacts
- [ ] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance


.eslintrc.js
```diff
@@ -20,7 +20,7 @@ module.exports = {
     'prettier/@typescript-eslint'
   ],
   rules: {
-    '@typescript-eslint/no-floating-promises': 'warn',
+    '@typescript-eslint/no-floating-promises': 'error',
```
mdurling: Is `error` the default? Can these rules be removed?

packages/mds-api-server/tests/index.spec.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable @typescript-eslint/no-floating-promises */
```
mdurling: So, `mocha` tests can be converted to use async/await instead of `done()` callbacks. Not worth it?

packages/mds-cache/index.ts
```diff
@@ -183,33 +183,32 @@ async function readDevicesStatus(query: { since?: number; skip?: number; take?:
...
-      device_status_map[item.device_id] = device_status_map[item.device_id] || {}
-      Object.assign(device_status_map[item.device_id], item)
-    })
-    log.info('readDevicesStatus', device_ids.length, 'entries:', all.length)
+  const device_ids_res = await client.zrangebyscoreAsync('device-ids', start, stop)
+  log.info('readDevicesStatus', device_ids_res.length, 'entries')
+
+  const skip = query.skip || 0
+  const take = query.take || 100000000000
+  const device_ids = device_ids_res.slice(skip, skip + take)
```
mdurling: Need to fix at some point ... too much mutation going on in this method.

packages/mds-api-server/tests/index.spec.ts
avatarneil: They can, but it's a pretty huge pain -- I think it makes more sense to just write tests using async/await moving forward. Much less crucial than having async/await for core packages IMO.

.eslintrc.js
karcass: pretty sure that this particular rule is `off` by default

PR #73, opened 2019-07-23
[mds-provider] Reinstate legacy provider endpoint handlers

mdurling:
Reinstated the legacy handlers for `/status_changes` and `/trips` ... this is temporary and will allow us to deploy updated versions of `mds-provider` until such time as the new event processing is ready.

## Impacts
- [x] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance


mdurling:
I can look into it, but there are two implementations of a two different endpoints. One is the current/legacy version that's being used and tested. The other is the new version that's not being used currently. The same tests apply to both. Once there are no longer two versions then the percentage can go back up, but I'll see what I can do.

mdurling:
> Can we add some basic tests, even though this is temporary? Reducing coverage by 5% makes me very sads 😿 (and we probably used to have tests for this stuff..?)

@avatarneil Coverage is back to its original level.

PR #74, opened 2019-07-24
[mds-stream] Updates to support knative eventing.

avatarneil:
## PR Checklist

## Impacts
- [ ] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance
- [x] Stream


packages/mds-stream/index.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable import/export */
```
mdurling: Why do we have to disable?

packages/mds-stream/index.ts
```diff
@@ -15,9 +16,36 @@
  */
 
 import logger from 'mds-logger'
+import { env } from 'process'
```
mdurling: Should be `const { env } = process`

packages/mds-stream/index.ts
```diff
@@ -135,15 +167,25 @@ async function writeStreamBatch(stream: Stream, field: string, values: unknown[]
 
 // put basics of vehicle in the cache
 async function writeDevice(device: Device) {
+  if (env.SINK) {
+    return write('ai.lacuna.device', JSON.stringify(device))
```
mdurling: That 6-character sequence of letters should not appear.

packages/mds-stream/index.ts
```diff
@@ -135,15 +167,25 @@ async function writeStreamBatch(stream: Stream, field: string, values: unknown[]
 
 // put basics of vehicle in the cache
 async function writeDevice(device: Device) {
+  if (env.SINK) {
+    return write('mds.device', JSON.stringify(device))
```
mdurling: Is there a reason that we can't use the same names? `DEVICE_INDEX_STREAM`?

packages/mds-stream/index.ts
```diff
@@ -135,15 +167,25 @@ async function writeStreamBatch(stream: Stream, field: string, values: unknown[]
...
+  }
   return writeStream(DEVICE_INDEX_STREAM, 'data', device)
 }
 
 async function writeEvent(event: VehicleEvent) {
+  if (env.SINK) {
+    return write('mds.event', JSON.stringify(event))
+  }
   return Promise.all([DEVICE_RAW_STREAM, PROVIDER_EVENT_STREAM].map(stream => writeStream(stream, 'event', event)))
```
mdurling: Notice that events get written to 2 different Redis streams, but only one knative. I am not sure how the `provider:event` stuff translates to knative.

packages/mds-stream/index.ts
```diff
@@ -17,8 +17,35 @@
...
+
+function getBinding() {
+  if (!binding) {
+    const config = {
+      method: 'POST',
+      url: env.SINK
+    }
+
+    // eslint-disable-next-line new-cap
+    return new Cloudevent.bindings['http-binary0.2'](config)
```
janedotx: Shouldn't there be a line like `binding = new Cloudevent.binds['http-binary0.2'](config)` to cache it?

avatarneil: Why not? Changed in 5c0527e

packages/mds-stream/index.ts
```diff
@@ -135,15 +168,25 @@ async function writeStreamBatch(stream: Stream, field: string, values: unknown[]
 
 // put basics of vehicle in the cache
 async function writeDevice(device: Device) {
+  if (env.SINK) {
+    return write('mds.device', JSON.stringify(device))
```
mdurling: I think `write` is a pretty generic name here. Should we use `emitCloudEvent` or something more narrative, similar to how we call it `writeStream`?

PR #82, opened 2019-07-25
[mds-db] Add indexed identity column to all tables

mdurling:
- Adds a DB generated indexed identity column to all tables. 
- All INSERTS return an updated object (with `id` and all default column values) for writing to cache/stream. 
- Use utils functions to simplify bulk inserts (`telemetry`, `status_changes`, `trips`).
- Enable latest audit migration.

## Impacts
- [x] Provider
- [x] Agency
- [x] Audit
- [x] Native
- [x] Policy
- [x] Compliance


packages/mds-db/migration.ts
```diff
@@ -189,12 +190,49 @@ async function recreateProviderTables(client: MDSPostgresClient) {
     }
   }
 }
+
+async function addIdentityColumnToAllTables(client: MDSPostgresClient) {
+  const exec = SqlExecuter(client)
+
+  const { rows: existing }: { rows: { table_name: string }[] } = await exec(
+    `SELECT table_name FROM information_schema.columns WHERE column_name = '${schema.IDENTITY_COLUMN}' AND table_catalog = CURRENT_CATALOG AND table_schema= CURRENT_SCHEMA`
```
avatarneil: super tiny nit -- there should be a space after `table_schema`

packages/mds-types/index.ts
```diff
@@ -349,9 +349,9 @@ export interface TripsStats {
 
 // The above types represent objects that can be created and passed into functions that write to the database. The
 // following type alias allows wrapping the above types with Recorded<> in order to represent what is read from the
-// database. This type alias will add the readonly attribute to all properties and also remove undefined as a valid
-// value since the database will never return undefined.
-export type Recorded<T> = Readonly<Required<T>>
+// database. This type alias will add the identity column, remove the readonly attribute from all properties, and
```
avatarneil: It seems like this adds the readonly attribute, no?

mdurling: No, it also adds the `id` column, etc. 
```
export type Recorded<T> = Readonly<Required<T & { id: number }>>
```

avatarneil: Yes, I see that. It _also_ *adds* the readonly attribute, however the comment dictates that it *removes* the readonly attribute. `remove the readonly attribute from all properties`. I'm saying you should update the comment to match what the behavior is

packages/mds-db/migration.ts
mdurling: Fixed in multiple places

packages/mds-db/sql-utils.ts
```diff
@@ -82,23 +82,32 @@ export function configureClient(pg_info: PGInfo) {
...
-export function vals_obj(cols: Readonly<string[]>, list: (string | number)[]): { [propName: string]: string | number } {
-  return range(cols.length).reduce((map, i: number) => Object.assign(map, { [i]: list[i] }), {})
+  return cols
+    .filter(col => col !== schema.IDENTITY_COLUMN)
+    .map(col => {
+      const value = obj[col]
+      if (typeof value === 'undefined') {
+        return null
+      }
+      if (value === null || typeof value === 'string' || typeof value === 'number') {
```
Levitikon217: Curious if there's a reason only strings and numbers are supported. No bools, dates, or other types?

mdurling: Until recently at least, our schema only used types that could be represented by javascript a javascript `string` (`varchar`, `uuid`) or `number` (`bigint`, `smallint`, `real`, etc.) ... we use timestamps for all _date_ columns and historically we have used a timestamp for things that can be treated as boolean values as well, but I will double check that this is sill true.

mdurling: Updated types and checking to be more representative of current types stored in db.

packages/mds-db/migration.ts
```diff
@@ -189,12 +190,53 @@ async function recreateProviderTables(client: MDSPostgresClient) {
...
 }
+
+async function addIdentityColumnToAllTables(client: MDSPostgresClient) {
+  const exec = SqlExecuter(client)
+
+  const { rows: existing }: { rows: { table_name: string }[] } = await exec(
+    `SELECT table_name FROM information_schema.columns WHERE column_name = '${schema.IDENTITY_COLUMN}' AND table_catalog = CURRENT_CATALOG AND table_schema = CURRENT_SCHEMA`
+  )
+
+  const create = Object.keys(schema.tables)
```
karcass: I do like programmatic DB maintenance

packages/mds-db/schema.ts
```diff
@@ -9,8 +9,11 @@ const AUDIT_EVENTS_TABLE = 'audit_events'
 const POLICIES_TABLE = 'policies'
 const GEOGRAPHIES_TABLE = 'geographies'
 
+const IDENTITY_COLUMN = 'id'
+
 // agency
 const DEVICES_COLS = [
+  IDENTITY_COLUMN,
   'device_id',
```
karcass: do we want constants for these other fields, since they are replicated?

packages/mds-utils/validators.ts
```diff
@@ -161,14 +161,19 @@ export const isValidAuditIssueCode = (
...
-  return event.event_type && event.telemetry
+export const isStringifiedEventWithTelemetry = (event: unknown): event is StringifiedEventWithTelemetry => {
+  return !!((event as StringifiedEventWithTelemetry).event_type && (event as StringifiedEventWithTelemetry).telemetry)
 }
 
-export const isStringifiedCacheReadDeviceResult = (device: any): device is StringifiedCacheReadDeviceResult => {
-  return device.device_id && device.provider_id && device.type && device.propulsion
+export const isStringifiedCacheReadDeviceResult = (device: unknown): device is StringifiedCacheReadDeviceResult => {
+  return !!(
+    (device as StringifiedCacheReadDeviceResult).device_id &&
```
karcass: is there a cleaner way to do this?  4x casts seems .... suboptimal?

mdurling: Type assertions inside TypeScript type guards are pretty commonplace. There is no runtime performance impact, but it does make it difficult to read in this instance. I'll see if there's a cleaner way.
 
> The reason why it's not called "type casting" is that casting generally implies some sort of runtime support. However, [type assertions](https://basarat.gitbooks.io/typescript/docs/types/type-assertion.html) are purely a compile time construct and a way for you to provide hints to the compiler on how you want your code to be analyzed.

packages/mds-db/schema.ts
mdurling: Yeah, I am going to tidy up the schema types to prevent typos and omissions.

avatarneil:
@mdurling could you merge in master? #98 _looks_ like it contains more changes than it actually does vs this branch.

mdurling:
> @mdurling could you merge in master? #98 _looks_ like it contains more changes than it actually does vs this branch.

Everything should be up to date with origin now.

PR #83, opened 2019-07-26
Update dependencies

mdurling:
Updating dependencies to latest versions. Also, switching back to `node 8.10.0` to match the runtime version. Outstanding question. Should we add a check to enforce this version as part of `test`?


avatarneil:
It'd be very nice if we added a check to enforce the version as part of `test` I think, yeah

mdurling:
> It'd be very nice if we added a check to enforce the version as part of `test` I think, yeah

Added node version check to test command.

PR #84, opened 2019-07-29
[mds-core] Add build info to /health response

mdurling:
Adding build version info to the `/health` response


karcass:
What's the plan for communicating changes to slack?  separate cron lambda?  something else?

PR #87, opened 2019-07-30
[mds-provider] Add back more legacy event types...

avatarneil:
## PR Checklist

## Impacts
- [x] Provider


packages/mds-provider/utils.ts
```diff
@@ -33,6 +33,27 @@ export function asStatusChangeEvent({ event_type }: VehicleEvent | { event_type:
...
+
+    case 'rebalance_drop_off':
+    case 'provider_drop_off':
+    case 'maintenance_drop_off':
+      return {
+        event_type: VEHICLE_STATUSES.available,
+        event_type_reason: VEHICLE_EVENTS.provider_drop_off
+      }
+
+    case 'provider_pick_up':
```
mdurling: `VEHICLE_EVENT.provider_pick_up` is in `case` below.

packages/mds-provider/utils.ts
```diff
@@ -33,6 +33,27 @@ export function asStatusChangeEvent({ event_type }: VehicleEvent | { event_type:
         event_type: VEHICLE_STATUSES.removed,
         event_type_reason: VEHICLE_EVENTS.service_end
       }
+
+    case 'rebalance_drop_off':
+    case 'provider_drop_off':
```
mdurling: `VEHICLE_EVENT.provider_drop_off` is in `case` below.

packages/mds-provider/utils.ts
```diff
@@ -33,6 +33,20 @@ export function asStatusChangeEvent({ event_type }: VehicleEvent | { event_type:
         event_type: VEHICLE_STATUSES.removed,
         event_type_reason: VEHICLE_EVENTS.service_end
       }
+
+    case 'rebalance_drop_off':
```
mdurling: Should move these two to the `VEHICLE_EVENTS.provider_drop_off` case ... comment below seems to indicate that's where they should go.

PR #90, opened 2019-08-01
FIx provider event_type_reason

mdurling:
Provider `event_type_reason` should be `user_pick_up` not `trip_start`

## Impacts
- [x] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance

### DIFF
--- a/packages/mds-provider/utils.ts
+++ b/packages/mds-provider/utils.ts
@@ -44,13 +44,13 @@ export function asStatusChangeEvent({ event_type }: VehicleEvent): StatusChangeE
     case VEHICLE_EVENTS.trip_end:
       return {
         event_type: VEHICLE_STATUSES.available,
-        event_type_reason: VEHICLE_EVENTS.user_drop_off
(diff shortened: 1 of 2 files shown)


packages/mds-provider/utils.ts
```diff
@@ -44,13 +44,13 @@ export function asStatusChangeEvent({ event_type }: VehicleEvent): StatusChangeE
...
       return {
         event_type: VEHICLE_STATUSES.available,
-        event_type_reason: VEHICLE_EVENTS.user_drop_off
+        event_type_reason: VEHICLE_REASONS.user_drop_off
       }
 
     case VEHICLE_EVENTS.provider_drop_off:
       return {
         event_type: VEHICLE_STATUSES.available,
-        event_type_reason: event_type // rebalance_drop_off or maintenance_drop_off
```
karcass: ideally we find some way to know if it's rebal or maint, but for now this is fine.

PR #91, opened 2019-08-01
[mds-api-server] API Maintenance Mode

mdurling:
Allow setting a `MAINTENANCE` environment variable to put an API in maintenance mode which will cause it to respond to all requests with `503 - Service Unavailable`

## Impacts
- [x] Provider
- [x] Agency
- [x] Audit
- [x] Policy
- [x] Compliance
- [x] Native

`MAINTENANCE=Database Migration` response to _ALL requests_:
Screen Shot 2019-08-01 at 8 19 08 PM

### DIFF
--- a/packages/mds-api-server/index.ts
+++ b/packages/mds-api-server/index.ts
@@ -42,18 +42,10 @@ export const ApiServer = (
   // Authorizer
   app.use((req: ApiRequest, res: ApiResponse, next: express.NextFunction) => {
     const { MAINTENANCE: maintenance } = process.env
-    res.locals.claims = authorizer(req)
     if (maintenance) {
-      if (req.method !== 'OPTIONS') {
-        if (
-          !pathsFor('/')
-            .concat(pathsFor('/health'))
-            .includes(req.path)
-        ) {
-          return res.status(503).send(about())
-        }
-      }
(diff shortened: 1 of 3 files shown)


packages/mds-api-server/index.ts
```diff
@@ -34,7 +41,19 @@ export const ApiServer = (
 ): express.Express => {
   // Authorizer
   app.use((req: ApiRequest, res: ApiResponse, next: express.NextFunction) => {
+    const { MAINTENANCE: maintenance } = process.env
     res.locals.claims = authorizer(req)
+    if (maintenance) {
+      if (req.method !== 'OPTIONS') {
+        if (
+          !pathsFor('/')
```
avatarneil: Make these also return 503's

mdurling: Done

packages/mds-api-server/index.ts
```diff
@@ -34,6 +41,10 @@ export const ApiServer = (
 ): express.Express => {
   // Authorizer
   app.use((req: ApiRequest, res: ApiResponse, next: express.NextFunction) => {
+    const { MAINTENANCE: maintenance } = process.env
```
karcass: deconstruction is so convenient

PR #94, opened 2019-08-02
[containers/k8s] add container image bundles and helm installer

invertigo:
Add new directory trees `container-images` and `helm` for building docker-style container images with yarn for each API, and for deploying said images to kubernetes with the helm package manager.

No impacts to existing functionality.  Adds a new build script `image` to the top-level `package.json`.


container-images/aws-utils/index.ts
```diff
@@ -0,0 +1,26 @@
+import KMS from 'aws-sdk/clients/kms'
```
mdurling: This looks like an exact copy of what's in `@aws-lambda/aws-utils` which raises some questions. 1) Do we even need it? 2) Is there a better way to handle secrets/encryption in a docker image than using KMS? 3) Can this module just be imported from somewhere common?

container-images/env-inject/index.ts
```diff
@@ -0,0 +1,32 @@
+/*
```
mdurling: This is cool. I am going to see if there might be a way to do this directly from within `webpack` so we wouldn't need this at all.

container-images/mds-agency/index.ts
```diff
@@ -0,0 +1,23 @@
+/*
```
mdurling: We can probably delete all of the `index.ts` files in `container-images` ... they are the AWS lambda handlers that won't be used.

container-images/mds-agency/server.ts
```diff
@@ -0,0 +1,29 @@
+/*
```
mdurling: If we delete all of the `index.ts` files then `server.ts` could become `index.ts`

container-images/mds-agency/tsconfig.build.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "../../tsconfig.build.json",
+  "compilerOptions": {
+    "outDir": "dist"
+  },
+  "references": [
+    { "path": "../../aws-lambda/aws-serverless-express-handler/tsconfig.build.json" },
```
mdurling: Again, I don't think we need to reference `aws-lambda/aws-serverless-express-handler` at all here.

container-images/mds-agency/webpack.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = (env, argv) => require('../webpack.config')({
+  env,
+  argv,
+  dirname: require('path').resolve(__dirname),
+  bundles: ['server']
```
mdurling: If we rename `index.ts` to `server.ts` then this would change to `['index']`

container-images/mds-audit/Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+RUN apk add --no-cache tini
+
+RUN mkdir /mds
+
+COPY dist/* /mds/
+
+WORKDIR /mds
+
+ENTRYPOINT ["/sbin/tini", "node", "server.js"]
```
mdurling: And this to `ENTRYPOINT ["/sbin/tini", "node", "server.js"]`

container-images/mds-metrics-sheet/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:8-alpine
```
mdurling: `mds-metrics-sheet` is a little different because it doesn't have a server component. They are just scheduled lambdas.

container-images/mds-provider/event-processor.ts
```diff
@@ -0,0 +1,27 @@
+/*
```
mdurling: I don't think we're quite ready for `event-processor` yet. I would remove this file.

container-images/mds-provider/webpack.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = (env, argv) => require('../webpack.config')({
+  env,
+  argv,
+  dirname: require('path').resolve(__dirname),
+  bundles: ['server', 'event-processor']
```
mdurling: And remove `event-processor` from here. We can add it back later if required.

container-images/aws-utils/index.ts
invertigo: I can get rid of this - kubernetes has its own secret storage which I've updated the helm chart to use

container-images/mds-metrics-sheet/Dockerfile
invertigo: Should run just fine as a cron job in k8s

mdurling: Sure, but there is no `server.ts` in this package presently.

invertigo: ah, gotcha.  is metrics-log.js the correct entry point?

mdurling: I think potentially there are 2 different entry points. Neil is probably most familiar.

invertigo: Updated the entrypoint to use the correct file.

container-images/mds-provider/event-processor.ts
invertigo: removed

container-images/webpack.config.js
```diff
@@ -0,0 +1,65 @@
+const webpack = require('webpack')
```
mdurling: If this is identical to `aws-lambda/webpack.config.js` we could probably just have one copy in the root folder and reference it from there via `require('../../webpack.config')`

invertigo: They're not quite the same - the container-images webpack config does not use the zip plugin.

mdurling: Fair. Would be pretty straight forward to make zip an option, but not much benefit ATM.

PR #96, opened 2019-08-05
[mds-compliance] Rework Count Rule Measurement

avatarneil:
Instead of allowing overflow within a rule's "bucket", force all overflow (vehicles which matched the rule\'s criteria but could not fit in the rule\'s bucket) to cascade down throughout the rule evaluation. At the end of the policy evaluation, all vehicles which have overflowed and never fit in a bucket will be considered violations.

## PR Checklist

## Impacts
- [ ] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [x] Compliance


avatarneil:
NOTE -- All external tools which depend on compliance output will need to be updated to parse the new output format (which is much simpler, now). Compliance spec will also need to be updated.

avatarneil:
Resolved #97

PR #101, opened 2019-08-08
[mds-core] Switch from Pre-Commit to using Husky and Talisman.

avatarneil:
Switch from pre-commit to using a combination of Husky and Talisman to ensure _no_ developers can check in secret keys.

## PR Checklist

## Impacts
- [ ] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance


.huskyrc.json
```diff
@@ -0,0 +1,6 @@
+{
+  "hooks": {
+    "pre-commit": "npm run husky-talisman -- pre-commit",
```
karcass: maybe a dumb quesiton, but why `npm` and not `yarn`?

package.json
```diff
@@ -42,6 +42,8 @@
     "eslint-plugin-prettier": "3.1.0",
     "eslint-plugin-promise": "4.2.1",
     "git-revision-webpack-plugin": "3.0.3",
+    "husky": "3.0.2",
```
mdurling: Cool. Does this mean we don't need the global install of `pre-commit` any longer?

avatarneil: Yup!

PR #106, opened 2019-08-13
[mds-metrics-sheet] Add `status` reporting to Google Metrics Sheet

avatarneil:
Add `status` specifications to the sheet, per the `event_type` -> `vehicle_status` in MDS.

### DIFF
--- a/packages/mds-metrics-sheet/metrics-log.ts
+++ b/packages/mds-metrics-sheet/metrics-log.ts
@@ -82,7 +82,10 @@ async function appendSheet(sheetName: string, rows: MetricsSheetRow[]) {
 function eventCountsToStatusCounts(events: { [s in VEHICLE_EVENT]: number }) {
   return Object.keys(events).reduce(
     (acc: { [s in VEHICLE_STATUS]: number }, event) => {
-      return Object.assign(acc, { [EVENT_STATUS_MAP[event as VEHICLE_EVENT]]: event })
+      return Object.assign(acc, {
+        [EVENT_STATUS_MAP[event as VEHICLE_EVENT]]:
+          acc[EVENT_STATUS_MAP[event as VEHICLE_EVENT]] + events[event as VEHICLE_EVENT]
+      })
     },
     {
       available: 0,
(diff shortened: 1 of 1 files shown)


packages/mds-metrics-sheet/metrics-log.ts
```diff
@@ -78,6 +79,23 @@ async function appendSheet(sheetName: string, rows: MetricsSheetRow[]) {
   }
 }
 
+function eventCountsToStatusCounts(events: { [s in VEHICLE_EVENT]: number }) {
+  return Object.keys(events).reduce(
+    (acc: { [s in VEHICLE_STATUS]: number }, event) => {
+      return Object.assign(acc, { [EVENT_STATUS_MAP[event as VEHICLE_EVENT]]: event })
```
mdurling: Should't this be an addition instead of assignment since multiple event types map to a single status?

avatarneil: Yep! Good catch 🤟🏼

packages/mds-metrics-sheet/metrics-log.ts
```diff
@@ -78,6 +79,26 @@ async function appendSheet(sheetName: string, rows: MetricsSheetRow[]) {
   }
 }
 
+function eventCountsToStatusCounts(events: { [s in VEHICLE_EVENT]: number }) {
+  return Object.keys(events).reduce(
+    (acc: { [s in VEHICLE_STATUS]: number }, event) => {
+      return Object.assign(acc, {
```
mdurling: A `const status = EVENT_STATUS_MAP[event as VEHICLE_EVENT]` wouldn't hurt readability.

packages/mds-metrics-sheet/metrics-log.ts
```diff
@@ -78,6 +79,26 @@ async function appendSheet(sheetName: string, rows: MetricsSheetRow[]) {
   }
 }
 
+function eventCountsToStatusCounts(events: { [s in VEHICLE_EVENT]: number }) {
+  return Object.keys(events).reduce(
```
mdurling: Starting with `(Object.keys(events) as VEHICLE_EVENT[]).reduce` would allow you to get rid of all the other `as VEHICLE_EVENT` casts.

PR #115, opened 2019-08-20
[mds-agency][mds-db] Fix vehicles endpoint to read from DB instead of cache

janedotx:
## PR Checklist

## Impacts
- [ ] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance

### DIFF
--- a/packages/mds-agency/api.ts
+++ b/packages/mds-agency/api.ts
@@ -547,7 +547,7 @@ function api(app: express.Express): express.Express {
       const response = await getVehicles(skip, take, url, provider_id, req.query)
       return res.status(200).send(response)
     } catch (err) {
-      await log.error('readDeviceIds fail', err)
+      await log.error('getVehicles fail', err)
       res.status(500).send(new ServerError())
     }
   })
--- a/packages/mds-agency/api.ts
+++ b/packages/mds-agency/api.ts
@@ -57,8 +57,6 @@ import {
 } from '@mds-core/mds-utils'
 import { AgencyApiRequest, AgencyApiResponse } from '@mds-core/mds-agency/types'
 
-import { CacheReadDeviceResult } from '@mds-core/mds-cache/types'
-
 function api(app: express.Express): express.Express {
   /**
    * Agency-specific middleware to extract provider_id into locals, do some logging, etc.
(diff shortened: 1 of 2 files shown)


packages/mds-agency/api.ts
```diff
@@ -484,7 +482,7 @@ function api(app: express.Express): express.Express {
       : rows
 
     const deviceIdSubset = deviceIdSuperset.slice(skip, skip + take).map(record => record.device_id)
-    const devices = (await cache.readDevices(deviceIdSubset)).reduce((acc: Device[], device: CacheReadDeviceResult) => {
+    const devices = (await db.readDeviceList(deviceIdSubset)).reduce((acc: Device[], device: Device) => {
```
mdurling: I think the issue here is going to be that the cache contains at least one additional property, `status`, that is not stored in the db.
```
hgetall device:13c6cb86-978c-4633-a830-411eab28e1f2:device
 1) "provider_id"
 2) "5f7114d1-4091-46ee-b492-e55875f7de00"
 3) "vehicle_id"
 4) "13c6cb86-978c-4633-a830-411eab28e1f2"
 5) "type"
 6) "scooter"
 7) "propulsion.0"
 8) "electric"
 9) "propulsion.1"
10) "human"
11) "year"
12) "2019"
13) "mfgr"
14) "ScootFast"
15) "model"
16) "T100"
17) "recorded"
18) "1566230943903"
19) "status"
20) "removed"
```

packages/mds-db/index.ts
```diff
@@ -267,6 +267,9 @@ async function readDevice(
 }
 
 async function readDeviceList(device_ids: UUID[]) {
+  if (device_ids.length === 0) {
```
mdurling: Good short circuit.

packages/mds-agency/api.ts
avatarneil: I think this might be a non-issue. `getVehicles` also fetches events from the cache, and if it can't find a corresponding event for a given `device_id`, then it sets `status = removed`.

mdurling: Awesome.

PR #116, opened 2019-08-20
[mds-cache] Switch a log.warn to log.info to avoid spamming slack for a cache miss

avatarneil:
## PR Checklist

### DIFF
--- a/.talismanrc
+++ b/.talismanrc
@@ -2,3 +2,6 @@ fileignoreconfig:
 - filename: yarn.lock
   checksum: 02a19fab10ddc762593b76e4aa5a4700cff5705a153a9acfe97467d2c9163068
   ignore_detectors: []
+- filename: packages/mds-cache/index.ts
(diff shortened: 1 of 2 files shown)


.talismanrc
```diff
@@ -1,4 +1,7 @@
 fileignoreconfig:
 - filename: yarn.lock
   checksum: 02a19fab10ddc762593b76e4aa5a4700cff5705a153a9acfe97467d2c9163068
-  ignore_detectors: []
\ No newline at end of file
+  ignore_detectors: []
```
mdurling: ¯\_(ツ)_/¯

packages/mds-cache/index.ts
```diff
@@ -268,7 +268,7 @@ async function updateProviderStats(suffix: string, device_id: UUID, timestamp: T
     )
   } catch (err) {
     const msg = `cannot updateProviderStats for unknown ${device_id}: ${err.message}`
-    await log.warn(msg)
+    await log.info(msg)
```
mdurling: We should implement some sort of log throttling/debouncing one day.

PR #119, opened 2019-08-21
[mds-cache] Instead of attempting to look up the provider_id, simply provider it as a function param.

avatarneil:
## PR Checklist

## Impacts
- [ ] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance

### DIFF
--- a/packages/mds-cache/index.ts
+++ b/packages/mds-cache/index.ts
@@ -252,15 +252,14 @@ async function readDevicesStatus(query: { since?: number; skip?: number; take?: 
   return values.filter((item: any) => item.telemetry)
 }
 
-// get the provider for a device
-async function getProviderId(device_id: UUID) {
-  return (await readDevice(device_id)).provider_id
-}
-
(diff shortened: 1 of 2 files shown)


packages/mds-cache/index.ts
```diff
@@ -252,15 +252,14 @@ async function readDevicesStatus(query: { since?: number; skip?: number; take?:
...
 }
 
-// get the provider for a device
-async function getProviderId(device_id: UUID) {
-  return (await readDevice(device_id)).provider_id
-}
-
 // initial set of stats are super-simple: last-written values for device, event, and telemetry
-async function updateProviderStats(suffix: string, device_id: UUID, timestamp: Timestamp | undefined | null) {
+async function updateProviderStats(
```
mdurling: I would probably suggest you rearrange the arguments to make `timestamp` (which is effectively optional) last. Maybe:
```
updateProviderStats(
  suffix: string,
  provider_id: UUID,
  device_id: UUID,
  timestamp: Timestamp | null = null
)
```

PR #121, opened 2019-08-22
[mds-compliance] Implement count minimum violations for mds-compliance.

janedotx:
## PR Checklist

## Impacts
- [ ] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance

### DIFF
--- a/packages/mds-compliance/mds-compliance-engine.ts
+++ b/packages/mds-compliance/mds-compliance-engine.ts
@@ -212,8 +212,9 @@ function processPolicy(
   devices: { [d: string]: Device }
 ): ComplianceResponse | undefined {
   if (isPolicyActive(policy)) {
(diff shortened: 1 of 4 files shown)


packages/mds-compliance/mds-compliance-engine.ts
```diff
@@ -212,8 +212,9 @@ function processPolicy(
   devices: { [d: string]: Device }
 ): ComplianceResponse | undefined {
   if (isPolicyActive(policy)) {
-    const vehiclesToFilter: MatchedVehicle[] = []
+    let total_violations = 0
```
avatarneil: This seems really side-effecty... what happens if you have a policy with both a count max rule and then a count min rule? If they should both have violations, then I think the violations in the min rule will completely overwrite the violations from the count max rule.

PR #125, opened 2019-08-25
[mds-db] Refactor migrations

mdurling:
All migrations have been applied to all existing database instances so the legacy migration code is being removed in favor of a more robust implementation that offers increased control and ensures that migrations are run exactly once even in the presence of multiple processes/connections.

Also, updated the function that is used to create foreign keys to be more generic and establish a better naming convention.

Finally, some minor cleanup of `mds-policy-author` config/scripts.


.talismanrc
```diff
@@ -1,6 +1,6 @@
 fileignoreconfig:
 - filename: packages/mds-db/sql-utils.ts
-  checksum: c5448443595c520b55184bfb67a4b4ee0f346d6b1864d8738268c52177c567a8
+  checksum: 532fb512aa6868689a76d3a49904c56e255b77c98abbe7e1b57e640ef344d4e6
```
wesleyfsmith: What's the story with this checksum?

mdurling: `talisman` is part of our `pre-commit` git hook that attempts to prevent the possibility of commiting secrets. Personally, I don't care for the way it works. It has the advantage of being installed as a dev dependency, but I'd love to find a better alternative.

PR #144, opened 2019-09-03
[mds-compliance] Fix compliance maximum 0 evaluation, and add test case to explicitly exercise that behavior. Thanks javascript <3

avatarneil:
PS: I hate javascript :) 
## PR Checklist

## Impacts
- [x] Compliance

### DIFF
--- a/packages/mds-compliance/mds-compliance-engine.ts
+++ b/packages/mds-compliance/mds-compliance-engine.ts
@@ -245,8 +245,8 @@ function processPolicy(
                     match_instance: MatchedVehicle,
                     i: number
                   ) => {
-                    const maximum =
-                      rule.maximum === null || rule.maximum === undefined ? Number.POSITIVE_INFINITY : rule.maximum
+                    // If the rule has a defined maximum, use it, even if 0
+                    const maximum = rule.maximum == null ? Number.POSITIVE_INFINITY : rule.maximum
                     if (maximum && i < maximum) {
                       if (overflowVehiclesMap[match_instance.device_id]) {
                         delete overflowVehiclesMap[match_instance.device_id]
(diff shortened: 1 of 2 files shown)


packages/mds-compliance/tests/api.spec.ts
```diff
@@ -1088,4 +1113,33 @@ describe('Tests Compliance API:', () => {
         })
     })
   })
+
+  describe('Verifies max 0 count policy', () => {
+    before('boop', async () => {
```
janedotx: you cannot name it boop

avatarneil: I mean, I _CAN_. Though I probably shouldn't 😜

packages/mds-compliance/mds-compliance-engine.ts
```diff
@@ -245,7 +245,8 @@ function processPolicy(
                     match_instance: MatchedVehicle,
                     i: number
                   ) => {
-                    const maximum = rule.maximum || Number.POSITIVE_INFINITY
+                    // If the rule has a defined maximum, use it, even if 0
+                    const maximum = rule.maximum == null ? Number.POSITIVE_INFINITY : rule.maximum
```
wesleyfsmith: So if they don't define a max the rule is to let them have an unlimited number of scooters in the area? I would that we would want to require them to set some sort of upper bound, but idk

avatarneil: This is primarily so we have a way to encode endless size buckets, for the case in which we might want to catch all vehicles and prevent further rule evaluation for them. Unfortunately there's no way to specify a POS_INF in JSON, so we take a count rule with a lack of a maximum as there being an endlessly sized bucket that vehicles can fall into during evaluation.

avatarneil: So, say for example I have two geographies A and B, where A is a subset of B (so like a bunch of valid drop off zones with no cap inside of a park which has greater regulations, for example). You could encode a policy that allows them to have as many vehicles as they want in the valid drop off zones `geography: A, max: null/undef` for the first rule, and then `geography: B, max: 0` for the second.

PR #146, opened 2019-09-04
[mds-db][mds-policy-author] API layer will generate policy_id if not set.

janedotx:
## PR Checklist

## Impacts
- [ ] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance


karcass:
Concur

avatarneil:
Agreed with above comments.

packages/mds-types/index.ts
```diff
@@ -29,6 +29,8 @@ export type VEHICLE_TYPE = keyof typeof VEHICLE_TYPES
 export const RULE_TYPES = Enum('count', 'speed', 'time')
 export type RULE_TYPE = keyof typeof RULE_TYPES
 
+export type PartialBy<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>
```
mdurling: If you move the uuid generation to the API ... this oddly named type won't be necessary I think.

janedotx: done.

mdurling: This type isn't used. Remove.

packages/mds-policy-author/api.ts
```diff
@@ -188,6 +189,10 @@ function api(app: express.Express): express.Express {
 
   app.post(pathsFor('/policies'), async (req, res) => {
     const policy = req.body
+    if (!isUUID(policy.policy_id)) {
```
mdurling: So, if the body contains an invalid (non-UUID) `policy_id` value then this will overwrite it. I think supplying an invalid API should be an error.

packages/mds-policy-author/api.ts
```diff
@@ -188,6 +189,10 @@ function api(app: express.Express): express.Express {
...
+      /* eslint-disable-next-line no-param-reassign */
+      policy.policy_id = uuid.v4()
```
mdurling: This is the _only_ place in the code that uses `.v4()`; everywhere else just uses `uuid()`

packages/mds-policy-author/api.ts
```diff
@@ -188,6 +189,10 @@ function api(app: express.Express): express.Express {
 
   app.post(pathsFor('/policies'), async (req, res) => {
     const policy = req.body
```
mdurling: You can get rid of the `parm-reassign` and handle the default with something like this:
```
const policy = { policy_id: uuid(), ...req.body }
```

packages/mds-policy-author/api.ts
janedotx: Ok, so do you think that the ID generation for a new policy should be the client's responsibility? The reason I cut this PR was because it was unclear whether the `policy_id` would be mandatory from the client or not. I'm pretty indifferent, with a slight preference for the API or DB layer supplying a valid ID if none was given (like normally it's not the client's job to supply a unique ID, when I order something on Amazon I don't pick an ID for that order), but perhaps @avatarneil or @karcass has a stronger opinion.

mdurling: I don't think the client should _have to_ supply a unique ID, but if they _choose to_ supply one then it ought to be valid or the API should return an error.

janedotx: Ah, okay, that finally clicked.

packages/mds-types/index.ts
avatarneil: It's such a snazzy type, though!

mdurling: I understand what it does, but it's name, `PartialBy` is non-intuitive. It feels like `Optional<T, K extends keyof T>` or `WithOptionalProperties<T, K extends keyof T>` would be more accurate. Nonetheless, we don't use it :-) Nice to see whoever did this leveraging mapped types though!

packages/mds-policy-author/api.ts
janedotx: Cool, I didn't know you could do that.

PR #156, opened 2019-09-10
Convert numeric query string parameters before type guarding

mdurling:
Treating strings as numbers can lead to unpredictable results. 

## Impacts
- [ ] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance
- [ ] Daily
- [X] Native
- [ ] Policy Author

### DIFF
--- a/packages/mds-native/api.ts
+++ b/packages/mds-native/api.ts
@@ -84,17 +84,14 @@ function api(app: express.Express): express.Express {
       params: { cursor },
       query: { limit: query_limit, ...filters }
     } = req
-    const limit = numericQueryStringParam(query_limit)
+    const limit = numericQueryStringParam(query_limit) || 1000
     isValidNumber(limit, { required: false, min: 1, max: 1000, property: 'limit' })
     if (cursor) {
       if (Object.keys(filters).length > 0) {
         throw new ValidationError('unexpected_filters', { cursor, filters })
       }
       try {
-        return {
-          cursor: JSON.parse(Buffer.from(cursor, 'base64').toString('ascii')),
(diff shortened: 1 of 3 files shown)


packages/mds-native/api.ts
```diff
@@ -75,13 +75,16 @@ function api(app: express.Express): express.Express {
     last_id: number
   }>
 
+  const numericQueryStringParam = (param: string | undefined): number | undefined => (param ? Number(param) : undefined)
```
avatarneil: Might make sense to move this to `mds-utils`, I think we're doing similar things in other places.

mdurling: There may be good reason to have a `converters.ts` file in `mds-utils` that hosts various type conversions (flattening telemetry for example might fall into that category, for example). I don't want to hold up this fix for that refactor though. This is the only place that is passing query string parameters into a typeguard/validator at the moment. Most other places are using the validators against JSON request bodies that are already converted.

packages/mds-native/api.ts
```diff
@@ -75,13 +75,16 @@ function api(app: express.Express): express.Express {
...
+
   const getRequestParameters = (
     req: NativeApiGetEventsRequest
   ): { cursor: NativeApiGetEventsCursor; limit: number } => {
     const {
       params: { cursor },
-      query: { limit = 1000, ...filters }
+      query: { limit: query_limit, ...filters }
```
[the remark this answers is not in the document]
mdurling: They always could ... limit is still 1000 (see validator) ... default is set below.

PR #161, opened 2019-09-12
Express scope validation middleware; use to validate admin:all

mdurling:
Updated the Express middleware for checking access token scopes. Rewrote it to be more functional in order to support complex scope requirements using javascript expressions, for example:

```
checkScope(check => (check('scope:1') || check('scope:2')) && check('scope:3'))
```

Which will return `403 Forbidden` if the token scope claims don't satisfy the expression.

## Impacts
- [ ] Provider
- [x] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance
- [x] Daily
- [ ] Native
- [ ] Policy Author


avatarneil (approved):
Awesome. This is going to be extremely helpful once we start adding more scopes!

PR #162, opened 2019-09-16
[mds-audit] Enable scope enforcement for MDS Audit endpoints

mdurling:
This PR turns on scope enforcement for the audit endpoints.

## Impacts
- [ ] Provider
- [ ] Agency
- [x] Audit
- [ ] Policy
- [ ] Compliance
- [ ] Daily
- [ ] Native
- [ ] Policy Author


packages/mds-audit/api.ts
```diff
@@ -153,81 +154,85 @@ function api(app: express.Express): express.Express {
...
-          await writeAuditEvent({
-            audit_trip_id,
-            audit_event_id,
-            audit_subject_id,
-            audit_event_type: AUDIT_EVENT_TYPES.start,
-            ...flattenTelemetry(telemetry),
-            timestamp,
-            recorded
-          })
+  app.post(
```
janedotx: Sorry, I'm probably just bad at reading the Express documentation, but does each of these `app.BLAH` methods take an arbitrary number of arguments? Does it call each of them one after another?

avatarneil: See `Route Handlers` on https://expressjs.com/en/guide/routing.html
```
var cb0 = function (req, res, next) {
  console.log('CB0')
  next()
}

var cb1 = function (req, res, next) {
  console.log('CB1')
  next()
}

var cb2 = function (req, res) {
  res.send('Hello from C!')
}

app.get('/example/c', [cb0, cb1, cb2])
```

packages/mds-provider/tests/status-changes.spec.ts
```diff
@@ -113,6 +113,20 @@ const fixtures: {
 ]
 
 describe('Testing Mapping of Agency Events to Provider Status Changes', () => {
+  before(done => {
+    fixtures.forEach(({ agency, provider }) =>
+      // eslint-disable-next-line no-console
+      console.log(
```
avatarneil: Does this need to be a `console.log`? I feel like this would make more sense as a unit test if it was using `assert`, I don't expect this to ever throw an error.

mdurling: It's not intended to be any sort of a test ... I have been asked enough times to generate the mappings that I wanted an easy/permanent way to dump them out. I did something similar with the MDS API scope values/descriptions.

packages/mds-audit/api.ts
mdurling: Also, the array syntax is optional. `app.get('/example/c', cb0, cb1, cb2)` or `app.get('/example/c/', [cb0, cb1], cb2)` all work identically. The key differentiator is that the inner functions are middleware functions that need to call `next()` in order to pass control to subsequent middleware functions and ultimately the route handler.

janedotx: Ah thank you, that is very enlightening.

PR #167, opened 2019-09-18
[mds-metrics-sheet] Adds optional type to metrics event_counts_last_24h

macsj200:
This should squash this lambda log error:

2019-09-18T07:00:45.559Z    0c7fca82-d644-4985-8b37-15cbcd440f5e    ERROR    ERROR getProviderMetrics TypeError: Cannot read property 'trip_enter' of undefined

Testing plan:
1. Run test suite

Fixes MDSAAS-158

## PR Checklist

 - [ X] simple searchable title - `[mds-db] Add PG env var`, `[config] Fix eslint config`
 - [ X] briefly describe the changes in this PR
 - [X ] mark as draft if should not be merged

## Impacts
- [ ] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance
- [ ] Daily
- [ ] Native
- [ ] Policy Author

Not sure what this touches. Let me know if you have any suggestions for adding to the test suite.

### DIFF
--- a/packages/mds-metrics-sheet/metrics-log.ts
+++ b/packages/mds-metrics-sheet/metrics-log.ts
@@ -30,7 +30,7 @@ import {
   BOLT_PROVIDER_ID
 } from '@mds-core/mds-providers'
 import { VEHICLE_EVENT, EVENT_STATUS_MAP, VEHICLE_STATUS } from '@mds-core/mds-types'
-import { VehicleCountResponse, LastDayStatsResponse, MetricsSheetRow } from './types'
+import { VehicleCountResponse, LastDayStatsResponse, MetricsSheetRow, VehicleCountRow } from './types'
 
 // The list of providers ids on which to report
 const reportProviders = [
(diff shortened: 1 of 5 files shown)


packages/mds-metrics-sheet/package.json
```diff
@@ -19,7 +19,9 @@
     "@mds-core/mds-logger": "0.1.9",
     "@mds-core/mds-providers": "0.1.9",
     "@mds-core/mds-types": "0.1.6",
+    "@types/lodash": "4.14.138",
```
mdurling: This should be removed and re-installed as a dev dependency in the root package.json `yarn add -D -W @types/lodash`

packages/mds-metrics-sheet/metrics-log.ts
```diff
@@ -170,8 +172,8 @@ async function getProviderMetrics(iter: number): Promise<MetricsSheetRow[]> {
           providerdropoff: event_counts.provider_drop_off || 0,
           tripstart: starts,
           tripend: ends,
-          tripenter: last[provider.provider_id].event_counts_last_24h.trip_enter || 0,
-          tripleave: last[provider.provider_id].event_counts_last_24h.trip_leave || 0,
+          tripenter: get(event_counts_last_24h, 'trip_enter', 0),
```
mdurling: Not super clear why `lodash.get` is required here.

avatarneil: Agreed, I think this may be unnecessary now that we do the check for `event_counts_last_24h`. The previous logic of `event_counts_last_24h.trip_enter || 0` should still work. The initial bug was that we were attempting to get a property from an undefined element, but we're doing that no longer!

packages/mds-metrics-sheet/metrics-log.ts
```diff
@@ -143,11 +144,12 @@ async function getProviderMetrics(iter: number): Promise<MetricsSheetRow[]> {
...
-          event_counts = last[provider.provider_id].event_counts_last_24h
-          status_counts = eventCountsToStatusCounts(last[provider.provider_id].event_counts_last_24h)
-          starts = last[provider.provider_id].event_counts_last_24h.trip_start || 0
-          ends = last[provider.provider_id].event_counts_last_24h.trip_end || 0
+        const { event_counts_last_24h } = last[provider.provider_id]
+        if (event_counts_last_24h) {
+          event_counts = event_counts_last_24h
+          status_counts = eventCountsToStatusCounts(event_counts_last_24h)
+          starts = event_counts_last_24h.trip_start || 0
+          ends = event_counts_last_24h.trip_end || 0
```
avatarneil: This is where you'd want to extract the `trip_enter`s I think, avoids the need for lodash. Then simply stick that var in (potentially an undefined var) in the return statement!

macsj200:
WIP, don't merge yet

PR #171, opened 2019-09-20
[mds-db] New database select methods

mdurling:
Borrowing a little inspiration from Entity Framework I created some new database methods for being able to specify shape/type and cardinality when selecting data. I updated a couple of the Audit related queries to illustrate how this might be used. Feedback welcomed.

## Impacts
- [ ] Provider
- [ ] Agency
- [x] Audit
- [ ] Policy
- [ ] Compliance
- [ ] Daily
- [ ] Native
- [ ] Policy Author

### DIFF
--- a/packages/mds-db/sql-utils.ts
+++ b/packages/mds-db/sql-utils.ts
@@ -67,10 +67,14 @@ export class MDSPostgresClient extends PostgresClient {
   }
 
   // Runs a query and returns the selected rows as an array of objects of type R.
-  public async select<R extends QueryResultRow>(command: string, values: (string | number)[] = []) {
+  private async select<R extends QueryResultRow>(command: string, values: (string | number)[]) {
     await logSql(command, values)
     const result: QueryResult<R> = await this.query(command, values)
     return result.rows
+  }
+
+  public async selectAll<R extends QueryResultRow>(command: string, values: (string | number)[] = []) {
+    return this.select<R>(command, values)
   }
 
   // Runs a query and returns a single selected row as an object of type R. An exception
--- a/packages/mds-db/sql-utils.ts
+++ b/packages/mds-db/sql-utils.ts
@@ -66,40 +66,49 @@ export class MDSPostgresClient extends PostgresClient {
     this.connected = connected
   }
 
-  public reader<R extends QueryResultRow>(command: string, values: (string | number)[] = []) {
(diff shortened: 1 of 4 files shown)


janedotx:
I gotta fly out but I'll take a longer look later. In the meantime, this seems to be an improvement over what we're doing now, but I really think we should move to an ORM in the future and away from building our own SQL queries as much as possible.

karcass:
> we should move to an ORM in the future and away from building our own SQL queries as much as possible.

Agree!  The (ahem) artisanal SQL was a go-as-fast-as-possible shortcut dating back to the original prototype.

mdurling:
> > we should move to an ORM in the future and away from building our own SQL queries as much as possible.
> 
> Agree! The (ahem) artisanal SQL was a go-as-fast-as-possible shortcut dating back to the original prototype.

I don't disagree with either of you. I proposed this as a draft because I felt like this syntax was both more flexible while at the same time being less verbose and therefore perhaps somewhat less error prone. We can consider this an interim step or we can jettison it altogether; either way I found crafting it to be a useful exercise.

packages/mds-db/sql-utils.ts
```diff
@@ -46,6 +65,55 @@ export class MDSPostgresClient extends PostgresClient {
...
+    return result.rows
+  }
+
+  public async selectAll<R extends QueryResultRow>(command: string, values: (string | number)[] = []) {
+    return this.select<R>(command, values)
+  }
+
+  // Runs a query and returns a single selected row as an object of type R. An exception
+  // is thrown if there are no selected rows or more than one selected row.
+  public async selectOne<R extends QueryResultRow>(command: string, values: (string | number)[] = []) {
```
janedotx: Just curious, I thought you were mostly against throwing an exception if something was not found? Having to remember whether to use `selectOne` or `selectOneOrNull` seems like unnecessary mental overhead.

mdurling: I'm mostly against not having the choice. There's not much overhead in having the ability to ask for one where you expect there to be exactly one and only one vs when there may be one or none. Anyone familiar with Entity Framework will recognize the semantics.

janedotx (approved):

mdurling:
Deferring. Looking into ORM integration.

PR #173, opened 2019-09-21
Add CORS middleware

mdurling:
## Impacts
- [ ] Provider
- [ ] Agency
- [x] Audit
- [ ] Policy
- [ ] Compliance
- [ ] Daily
- [ ] Native
- [x] Policy Author

### DIFF
--- a/aws-lambda/aws-serverless-express-handler/index.ts
+++ b/aws-lambda/aws-serverless-express-handler/index.ts
@@ -61,7 +61,7 @@ export const AwsServerlessExpressHandler = (
   })
 
   const server = awsServerlessExpress.createServer(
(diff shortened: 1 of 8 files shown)


avatarneil (approved):

avatarneil:
This does technically impact _all_ the APIs, not just Audit and Policy-Author.

mdurling:
> This does technically impact _all_ the APIs, not just Audit and Policy-Author.

True, although Audit/Policy-Author are the only APIs that accepts requests from a browser. If `istio` handles CORS then we may not need this middleware at all.

mdurling:
> This does technically impact _all_ the APIs, not just Audit and Policy-Author.

@avatarneil I made `CORS` middleware opt-in. It doesn't need to be enabled when running on `API Gateway` or `Istio` ... so it is only enabled for web accessible APIs (audit, policy-author) when running locally.

PR #174, opened 2019-09-24
operational support enabling MDS deployments atop kubernetes

jwtodd:
Add operational support enabling kubernetes deployments, namely:

  * yarn docker image build support
  * helm/mds charts
  * supporting sh scripts

## PR Checklist

## Impacts
- [x] Provider
- [x] Agency
- [x] Audit
- [x] Policy
- [x] Compliance
- [x] Daily
- [x] Native
- [x] Policy Author


mdurling:
In general, `mds-policy-author` does not appear to be fully represented in this PR. There are several places where it appears to be missing. There is also another existing PR that perhaps needs to be merged/included here.

bin/install
```diff
@@ -0,0 +1,719 @@
...
+  ["istio.io"]="https://storage.googleapis.com/istio-release/releases/1.3.0/charts"
+  ["banzaicloud-stable"]="https://kubernetes-charts.banzaicloud.com"
+  ["elastic"]="https://helm.elastic.co")
+declare -A apps=(
+  ["mds-agency"]="https://localhost:[port]/agency,mds,app"
+  ["mds-audit"]="https://localhost:[port]/audit,mds,app"
+  ["mds-compliance"]="https://localhost:[port]/compliance,mds,app"
+  ["mds-daily"]="https://localhost:[port]/daily,mds,app"
+  ["mds-native"]="https://localhost:[port]/native,mds,app"
+  ["mds-policy"]="https://localhost:[port]/provider,mds,app"
```
mdurling: `mds-policy-author` missing?

helm/mds/templates/auth.yaml
```diff
@@ -0,0 +1,28 @@
...
+  origins:
+  - jwt:
+      audiences:
+      {{- range $.Values.jwt.audiences }}
+      - {{ . | quote }}
+      {{- end }}
+      issuer: {{ $.Values.jwt.issuer | quote }}
+      jwksUri: {{ printf "%s.well-known/jwks.json" $.Values.jwt.issuer }}
+      trigger_rules:
+      - excluded_paths:
```
mdurling: Should the `root` path be excluded?

helm/mds/templates/deployment.yaml
```diff
@@ -0,0 +1,102 @@
...
+  selector:
+    matchLabels:
+      app: {{ $api.name }}
+  template:
+    metadata:
+      labels:
+        app: {{ $api.name }}
+    spec:
+      containers:
+      - name: user-container
```
mdurling: The container name recently changed on sandbox. It used to be `user-container`. Is this PR missing some changes that are currently deployed? This is current from Kibana.
Screen Shot 2019-09-24 at 12 37 46 PM

helm/mds/tests/auth-agency-jwt-audience_test.yaml
```diff
@@ -0,0 +1,73 @@
...
+  - auth.yaml
+tests:
+  - it: doc 1
+    release:
+      namespace: mds
+    set:
+      apis:
+        - name: mds-provider
+        - name: mds-agency
+          pathPrefix: /agency
```
mdurling: Why is `pathPrefix` specified for `mds-agency` but not for other APIs?

helm/mds/tests/auth-agency-jwt-audience_test.yaml
```diff
@@ -0,0 +1,73 @@
+suite: test authetication jwt audience agency
```
mdurling: This is just a little bit of repo hygiene, but it appears all of the `helm` tests are located under the `helm/mds/tests` folder and the inclusion of `_test` in the file names is inconsistent. I'd recommend making the filenames consistent by either adding the `_test` (or preferably `.test`) to _all_ filenames or removing it.

helm/mds/templates/deployment.yaml
invertigo: Updated to account for this

helm/mds/templates/auth.yaml
invertigo: I don't think so - it isn't in AWS API Gateway

bin/install
jwtodd: resolved thx

helm/mds/tests/auth-agency-jwt-audience_test.yaml
jwtodd: these is a configuration overrlay/override whereby we are only testing a conditional agency deployment and as such only need to override the mds-agency configuration data, and because the apis are structured as a list, vs keyed map, we have to include the entire list when doing so, even stubs :(

helm/mds/tests/auth-agency-jwt-audience_test.yaml
jwtodd: i know helm-unittest [ https://github.com/lrills/helm-unittest ] recommends this naming convention. i am not sure if they require it. will test in time ...  assuming this is not a blocker at present. i would love to group the tests as suites as well, if only as distinct dirs. have yet to get that far.

PR #175, opened 2019-09-25
[mds-policy-author] Added scoping for policy author.

janedotx:
## PR Checklist

## Impacts
- [ ] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance
- [ ] Daily
- [ ] Native
- [ ] Policy Author


packages/mds-policy-author/tests/api.spec.ts
```diff
@@ -54,11 +55,8 @@ const request = supertest(ApiServer(api))
 
 const APP_JSON = 'application/json; charset=utf-8'
 
-// TODO
-// change the auth token/authing system so it uses agency_id instead of provider_id
-const AUTH_NON_PROVIDER = `basic ${Buffer.from(`'BOGUS_PROVIDER_ID_TO_BE_REPLACED'`).toString('base64')}`
-const AUTH_PROVIDER_ONLY = `basic ${Buffer.from(`${TEST1_PROVIDER_ID}`).toString('base64')}`
-const AUTH_BAD_PROVIDER_ONLY = `basic ${Buffer.from(`${TEST3_PROVIDER_ID}`).toString('base64')}`
+const SCOPED_AUTH = (...scopes: AccessTokenScope[]) =>
```
janedotx: Any thoughts on what a better place to put this might be? Maybe `mds-test-data`?

packages/mds-policy-author/api.ts
```diff
@@ -108,49 +106,7 @@ const featureCollectionSchema = Joi.object()
   .unknown(true) // TODO
 
 function api(app: express.Express): express.Express {
-  /**
```
mdurling: I guess you determined this `provider_id` check/filter is no longer necessary?

packages/mds-policy-author/tests/api.spec.ts
mdurling: That would require `mds-test-data` to take a dependency on `mds-api-scopes` which could get weird. Perhaps the definition of `AccessTokenScope` could be moved to `mds-types` and then moving `SCOPED_AUTH` to `mds-test-data` would make sense. Otherwise, it's going to continue to get replicated in every API project.

packages/mds-policy-author/package.json
```diff
@@ -12,6 +12,8 @@
   "dependencies": {
     "@hapi/joi": "15.1.1",
     "@mds-core/mds-api-server": "0.1.9",
+    "@mds-core/mds-api-helpers": "0.1.9",
+    "@mds-core/mds-api-server": "0.1.9",
```
mdurling: `mds-api-server` is in here twice now.

packages/mds-policy-author/tests/api.spec.ts
```diff
@@ -30,7 +30,8 @@ import db from '@mds-core/mds-db'
 import { clone } from '@mds-core/mds-utils'
 import { Policy } from '@mds-core/mds-types'
 import { ApiServer } from '@mds-core/mds-api-server'
-import { TEST1_PROVIDER_ID, TEST3_PROVIDER_ID } from '@mds-core/mds-providers'
+import { AccessTokenScope } from '@mds-core/mds-api-scopes'
```
mdurling: `mds-api-scopes` is not declared as a dependency in `package.json` or a project reference in `tsconfig.build.json`

packages/mds-policy-author/api.ts
janedotx: I thought we had this hack in just so Blue could log in, in the pre-auth0 world, but after discussing with @avatarneil , I'll put back in the `provider_id` checks to verify that the UUID is present and valid, but not the hack at the end.

mdurling: I don't know what the `hack` is that you are referring to, but my question should probably have been _"Does the access token need to have a valid provider id to access this API?"_ 

I think some applications may proxy requests using a M2M token containing a provider_id claim, but if other applications are doing user-level authentication then there is no mechanism for placing a provider_id in the token presently.

janedotx: @avatarneil actually thinks it's good to rip it all out.

packages/mds-policy-author/tests/api.spec.ts
janedotx: Thanks

packages/mds-policy-author/package.json
janedotx: fixed

packages/mds-policy-author/api.ts
avatarneil: Yeah, I agree it's all good to rip out. We definitely don't want a `provider_id` claim to be required for `policy-author` moving forward.

packages/mds-policy-author/api.ts
```diff
@@ -108,49 +106,7 @@ const featureCollectionSchema = Joi.object()
...
-      }
-    } catch (err) {
-      /* istanbul ignore next */
-      await log.error(req.originalUrl, 'request validation fail:', err.stack)
-    }
-    next()
-  })
-
-  app.get(pathsFor('/policies'), async (req, res) => {
+  app.get(pathsFor('/policies'), checkScope(check => check('policies:read')), async (req, res) => {
```
avatarneil: What would the scope be for a provider to read the policies that are applicable to them? `policies:read:provider`?

avatarneil: This isn't currently in the spec at all, but it's something I think needs to definitely be considered.

mdurling: Yes, that's the intent of `policies:read:provider` ... its enforcement would be a little different than just `checkScope` though.

packages/mds-test-data/index.ts
```diff
@@ -552,6 +553,9 @@ function makeTrip(device: Device): Trip {
   }
 }
 
+const SCOPED_AUTH = (scopes: AccessTokenScope[], provider_id = TEST1_PROVIDER_ID) =>
```
mdurling: Since this is not _always_ a `provider_id`, perhaps call this argument `principalId` to match how it is parsed in `mds-api-authorizer`

janedotx: 👍

PR #179, opened 2019-09-26
Assign Blue Systems a unique provider_id

mdurling:
Blue System was using incorrect and unofficial provider IDs. Reassigned them one of our test IDs to ensure that there are no issues during deployments and Auth0 updates.


packages/mds-compliance/api.ts
```diff
@@ -105,16 +118,14 @@ function api(app: express.Express): express.Express {
         policy &&
         ((policy.provider_ids && policy.provider_ids.includes(provider_id)) ||
           !policy.provider_ids ||
-          ([TEST1_PROVIDER_ID, TEST2_PROVIDER_ID, TEST4_PROVIDER_ID].includes(provider_id) &&
+          (AllowedProviderIDs.includes(provider_id) &&
```
janedotx: Not a blocker, but I'm wondering why the auth0 scoping can't be used here?

mdurling: Good question. Ultimately, that's the plan. We can't enable scope enforcement though until all providers switch over to using the new access token audiences.

packages/mds-providers/index.ts
```diff
@@ -177,17 +192,26 @@ export const providers: Readonly<{ [P in PROVIDER_ID]: Readonly<Provider> }> = O
...
+  [SUPPORT3_PROVIDER_ID]: Object.freeze({
+    provider_id: SUPPORT3_PROVIDER_ID,
+    provider_name: 'Support 3'
   }),
-  [TEST5_PROVIDER_ID]: Object.freeze({
-    provider_id: TEST5_PROVIDER_ID,
-    provider_name: 'Test 5'
+  [DEPRECATED_BLUE_SYSTEMS_PROVIDER_ID]: Object.freeze({
+    provider_id: BLUE_SYSTEMS_PROVIDER_ID,
+    provider_name: 'Blue Systems',
```
avatarneil: Might be worth making the `provider_name` `Blue Systems (DEPRECATED)` to make look-back easier

PR #180, opened 2019-09-26
helm values apis to keyed-map vs list for operational ease-of-use

jwtodd:
note: don't merge without consideration of production impact (namely downstream values.yaml file overlays will need to be modified)

reconsider changing the helm/mds/values.yaml apis structure from a list to a keyed map principally for ease of configuration overrides (eg: --set apis.mds-agency.migration=true)

@avatarneil @invertigo this is a significant helm-values structural change but in all likelihood a good best-practice given what we have learned with regards to having the operational ability to configure overrides, in that doing so against a keyed-map is trivial-and-consistent vs a list of which the order can change over time, applying a setter override against a list is an urban-myth (in that there is rumor as to means to do so but i have tried and i have failed).

i updated the tests accordingly:

  `% helm unittest ./helm/mds`

and as verification bin/install now defaults to mds-agency migration:true, eg:

  `["sets-mds"]="${MDS_SETS_MDS:-pgPass=Password123#,apis.mds-agency.migration=true}"`

in order to enable optional deployments, an mds 'enabled' attribute has been added that has a default value of true. it can be tested as follows:

`% ./bin/install ... -c:aps.mds-compliance.enabled=false ...`

additionally we can now add an optional list of env name-value tuples that will get inlined upon render:
 
```
  apis:
    foo:
      env:
        - name: foo
          value: bar
```
further, we can override default resources (ie cpu, memory) as follows:

`... . -c:sets-mds+=apis.resourcesLimitsCpu=100m,sets-mds+=apis.resourcesLimitsMemory=128Mi,sets-mds+=apis.resourcesRequestsCpu=10m,sets-mds+=apis.resourcesRequestsMemory=32Mi`

lastly, in re-looking into the helm-tests i noticed it is hard-if-not-impossible to test the non-existence of a configuration, specifically the helm/mds/tests/*-no-events_test.yaml tests, to which i have added FIXMEs

## PR Checklist

## Impacts
- [x] Provider
- [x] Agency
- [x] Audit
- [x] Policy
- [x] Compliance
- [x] Daily
- [x] Native
- [x] Policy Author


helm/mds/templates/deployment.yaml
```diff
@@ -1,37 +1,38 @@
...
           limits:
-            cpu: 500m
-            memory: 512Mi
+            cpu: {{ $.Values.resourcesLimitsCpu | default "500m" }}
+            memory: {{ $.Values.resourcesLimitsMemory | default "512Mi" }}
           requests:
-            cpu: 50m
-            memory: 128Mi
+            cpu: {{ $.Values.resourcesRequestsCpu | default "50m" }}
+            memory: {{ $.Values.resourcesRequestsMemory | default "128Mi" }}
```
invertigo: As long as we're doing this - I think it would be best to have the resources be set per-API rather than common across all APIs.  It's reasonable to expect that say mds-provider might have different requirements than mds-policy-author.

invertigo: So `{{ $api.resourcesLimitsCpu }}`, `{{ $api.resourceLimitsMemory }}`, etc.

oldschoolsysadmin (changes requested):
Looking good - one comment.  Also, this:

> `--set apis.mds-agency.migration=true`

Makes loads of sense!  That wasn't clear to me in our conversation the other day about the change, and I like how it brings the name of the API into the configuration namespace.  Good thinking.

helm/mds/templates/deployment.yaml
jwtodd: agreed. i have been looking at providing api-service-level resource overrides and ran into blockers given we are iterating over the api-services via a range function. turns out stepping away form the keyboard for a bit is "a good thing(tm)" and with a fresh approach i was able to push a change that provides, in order of precedence, specifying resource values at the api-service level, at the fleet (aka all services) and, lastly, the default (in template) value. eg:

```
# note: override mds-agency resources
% ... --set api.[api-service].[resourcesLimits[Cpu|Memory]=[X]|resourcesRequests[Cpu|Memory]=[Y]]
# note: override all service resources
% ... --set resourcesLimits[Cpu|Memory]=[X]|resourcesRequests[Cpu|Memory]=[Y]
```

note: these configurations can be used together, eg override all defaults while specifying distinct values per service

avatarneil (approved):

avatarneil:
@jwtodd Mind addressing merge conflicts?

jwtodd:
@avatarneil now mergable

@invertigo i believe the change-request to support varying resource configurations-per-service has been addressed? i am adding tests for this feature now.

oldschoolsysadmin (approved):
Looks good

PR #186, opened 2019-09-30
Upgrade dependencies

mdurling:
I updated the dependencies for everything not related specifically to the AWS Lambda deployment.

The latest versions introduced a few TS errors due to non-overlapping types that required a cast to unknown. @macsj200 Maybe there's a way to provide some test helper for constructing these req/res objects. Thoughts?


packages/mds-daily/tests/request-handlers.test.ts
```diff
@@ -14,9 +14,9 @@ describe('Request handlers', () => {
         } as any
       }
       await getRawTripData(
-        {
+        ({
           params: { trip_id: 'fake-trip-id' }
-        } as DailyApiRequest,
+        } as unknown) as DailyApiRequest,
```
janedotx: Just curious, why do we need this type annotation now?

mdurling: I regenerated the lock file and TS started complaining about this.
```
Conversion of type '{ params: { trip_id: string; }; }' to type 'Request' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Type '{ params: { trip_id: string; }; }' is missing the following properties from type 'Request': get, header, accepts, acceptsCharsets, and 66 more.
```
The recommended fix is to `Add 'unknown' conversion for non-overlapping types`

mdurling:
Not going to regenerate lock file.

PR #187, opened 2019-09-30
[mds-daily] Fix lambda error

macsj200:
This is an attempt to triage this error
`/admin/vehicle_counts fail TypeError: Cannot read property 'device_id' of null`

I haven't been able to pull specific log info yet, so it's just a guess as to the problem. Likely we're just running into a cache miss, so we're getting an unexpected `null` in the cache results, so I just added some branching logic to deal with that case.

## PR Checklist

 - [ X] simple searchable title - `[mds-db] Add PG env var`, `[config] Fix eslint config`
 - [ X] briefly describe the changes in this PR
 - [X ] write tests for all new functionality

## Impacts
- [ ] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance
- [ X] Daily
- [ ] Native
- [ ] Policy Author

### DIFF
--- a/packages/mds-daily/utils.ts
+++ b/packages/mds-daily/utils.ts
@@ -1,5 +1,7 @@
 import { isTimestamp, now, days, inc, head, tail } from '@mds-core/mds-utils'
-import { UUID, CountMap, TripsStats, VEHICLE_EVENTS } from '@mds-core/mds-types'
+import { UUID, CountMap, TripsStats, VEHICLE_EVENTS, VehicleEvent } from '@mds-core/mds-types'
+import cache from '@mds-core/mds-cache'
+import log from '@mds-core/mds-logger'
 
 // TODO move to utils?
 export function asInt(n: string | number | undefined): number | undefined {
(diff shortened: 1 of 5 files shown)


avatarneil (approved):
The error has ~magically disappeared~ whenever I try to hit the endpoint now... WTF? Good PR though, should be helpful in the case this decides to reappear.

packages/mds-daily/utils.ts
```diff
@@ -99,3 +101,29 @@ export function categorizeTrips(perTripId: TripsData): { [s: string]: TripsStats
   })
   return perProvider
 }
+
+export async function getMaps(): Promise<{
+  eventMap: { [s: string]: VehicleEvent }
+  // telemetryMap: { [s: string]: Telemetry }
+}> {
+  try {
+    // const telemetry: Telemetry[] = await cache.readAllTelemetry()
```
janedotx: mind removing this comment? thanks

packages/mds-daily/utils.ts
```diff
@@ -99,3 +101,29 @@ export function categorizeTrips(perTripId: TripsData): { [s: string]: TripsStats
...
+    // log.info('read telemetry')
+    const events = await cache.readAllEvents()
+    log.info('read events')
+    const eventSeed: { [s: string]: VehicleEvent } = {}
+    const eventMap: { [s: string]: VehicleEvent } = events.reduce((map, event) => {
+      return event ? Object.assign(map, { [event.device_id]: event }) : map
+    }, eventSeed)
+    // const telemetrySeed: { [s: string]: Telemetry } = {}
```
janedotx: these too

PR #191, opened 2019-10-01
Change checkScope to support additional use cases

mdurling:
Change the scope validator to be a function that gets a list of user scopes and returns a boolean to support additional validation expressions.

## Impacts
- [x] Provider
- [ ] Agency
- [x] Audit
- [ ] Policy
- [ ] Compliance
- [ ] Daily
- [x] Native
- [x] Policy Author


levi217:
@twelch FYI, this should support the 'user has ANY permission scenario'

PR #193, opened 2019-10-02
[mds-agency] Permit access to reading all /vehicles/ data if vehicles:read scope is present.

avatarneil:
## PR Checklist

## Impacts
- [ ] Provider
- [x] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance
- [ ] Daily
- [ ] Native
- [ ] Policy Author

### DIFF
--- a/packages/mds-agency/request-handlers.ts
+++ b/packages/mds-agency/request-handlers.ts
@@ -140,7 +140,7 @@ export const getVehicleById = async (req: AgencyApiRequest, res: AgencyApiRespon
   log.info(`/vehicles/${device_id}`, cached)
   const store = cached ? cache : db
   const payload = await readPayload(store, device_id)
-  if (!payload.device || payload.device.provider_id !== provider_id) {
+  if (!payload.device || (provider_id && payload.device.provider_id !== provider_id)) {
     res.status(404).send({
       error: 'not_found'
     })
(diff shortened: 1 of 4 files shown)


packages/mds-agency/request-handlers.ts
```diff
@@ -135,7 +135,7 @@ export const getVehicleById = async (req: AgencyApiRequest, res: AgencyApiRespon
 
   const { cached } = req.query
 
-  const { provider_id } = res.locals
+  const { provider_id } = res.locals.scopes.includes('vehicles:read') ? req.query : res.locals
```
mdurling: `GitHub` won't let me comment, but I think lines `142-148` need to change as well. Currently, if the request is scoped with `vehicles:read` then the request will return `404` _unless_ it also contains a matching _provider_id_ in the query string which seems unnecessary. The solution could be to adjust the `if` condition or perhaps pass an _optional_ provider_id to `readPayload`.

mdurling (changes requested):
See comment concerning `vehicles:read` edge case on `/vehicles/:device_id`

packages/mds-agency/request-handlers.ts
avatarneil: SGTM, fixed

mdurling (approved):

PR #194, opened 2019-10-02
[mds-compliance] Finalize implementation of speed rule compliance

janedotx:
## PR Checklist

## Impacts
- [ ] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [x] Compliance
- [ ] Daily
- [ ] Native
- [ ] Policy Author

### DIFF
--- a/packages/mds-compliance/mds-compliance-engine.ts
+++ b/packages/mds-compliance/mds-compliance-engine.ts
@@ -224,7 +224,10 @@ function processPolicy(
   if (isPolicyActive(policy)) {
     const vehiclesToFilter: MatchedVehicle[] = []
     let overflowVehiclesMap: { [key: string]: MatchedVehicle & { rule_id: UUID } } = {}
+    const speedingVehiclesMap: { [key: string]: MatchedVehicle & { rule_id: UUID } } = {}
+    const allSpeedingVehicles = []
     const compliance: Compliance[] = policy.rules.reduce((compliance_acc: Compliance[], rule: Rule): Compliance[] => {
+      // This is because even if a vehicle breaks two rules, it will be counted in violation of only the first rule.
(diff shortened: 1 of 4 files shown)


packages/mds-compliance/mds-compliance-engine.ts
```diff
@@ -304,7 +314,23 @@ function processPolicy(
...
+            compliance_acc.push(comp)
+            vehiclesToFilter.push(
+              ...(comp.matches
+                ? comp.matches.reduce((acc: MatchedVehicle[], match: SpeedMatch) => {
+                    acc.push(match.matched_vehicle)
+                    return acc
+                  }, [])
+                : [])
+            )
+          }
```
avatarneil: All of the vehicles which are `matched` in here need to be counted in the `total_violations` field on line 349, and included in the `vehicles_in_violation` section. We don't need the entire compliance output for each rule, you should do the minified version in the same style as `count` rules (which basically just reports the measured number for each geography_id in a rule)

packages/mds-compliance/tests/engine-test.spec.ts
```diff
@@ -140,38 +140,37 @@ describe('Tests Compliance Engine', () => {
...
+    const results = filteredPolicies.map(policy => processPolicy(policy, filteredEvents, geographies, deviceMap))
+    results.forEach(result => {
+      if (result) {
+        result.compliance.forEach(compliance => {
+          if (
+            compliance.rule.geographies.includes(CITY_OF_LA) &&
+            compliance.matches &&
+            compliance.rule.rule_type === RULE_TYPES.speed
+          ) {
+            test.assert(compliance.matches.length !== 0)
```
avatarneil: Test should be updated to check the `total_violations` param, and there should be 5 of them

PR #195, opened 2019-10-03
suffix the container image tag based on the current git branch

invertigo:
Moves the docker build command from `package.json` to a utility script, so that we can check the current branch and suffix the container image tag as needed.

## PR Checklist

## Impacts
- [ ] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance
- [ ] Daily
- [ ] Native
- [ ] Policy Author

### DIFF
--- a/bin/image-build
+++ b/bin/image-build
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+# Tag images built outside of the master branch with a
+# "-prerelease" suffix for deployment safety.
+
+IMAGE="$1"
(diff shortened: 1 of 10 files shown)


invertigo:
> LGTM. Does this work _as-is_ on MacOS, Win, linux or are there any special considerations?

The shell script is simple enough to work with OSX's ancient bash, and the docker commands don't use any os-specific features like volume mounts.  Should be fine.

PR #202, opened 2019-10-04
[mds-audit] Store latest provider event with audit events; add get vehicle endpoint

BigIdeas:
## PR Checklist

## Impacts
- [ ] Provider
- [ ] Agency
- [x] Audit
- [ ] Policy
- [ ] Compliance
- [ ] Daily
- [ ] Native
- [ ] Policy Author


packages/mds-audit/api.ts
```diff
@@ -116,14 +119,27 @@ function api(app: express.Express): express.Express {
...
+      if ((!provider_id || !provider_vehicle_id) && res.locals.audit) {
+        provider_id = res.locals.audit.provider_id
+        provider_vehicle_id = res.locals.audit.provider_vehicle_id
+      }
+      const device = await readDeviceByVehicleId(provider_id, provider_vehicle_id)
+      try {
+        const provider_event = await readEvent(device.device_id)
+        res.locals.provider_event = provider_event
+      } catch (err) {
+        await log.warn('Unable to find provider event to match with audit event', JSON.stringify(err))
```
avatarneil: I'm pretty sure the logger will unwrap any JSON you provide, so the `JSON.stringify()` might be unnecessary

packages/mds-audit/api.ts
```diff
@@ -188,12 +204,18 @@ function api(app: express.Express): express.Express {
             })
 
             // Create the audit start event
+            const provider_event_id = res.locals.provider_event && res.locals.provider_event.id
```
avatarneil: I am so looking forward to the `?.` operator in TS 3.7!!

packages/mds-audit/service.ts
```diff
@@ -105,6 +105,11 @@ export async function readDeviceByVehicleId(provider_id: UUID, vehicle_id: strin
   }
 }
 
+export async function readEvent(device_id: UUID): Promise<Recorded<VehicleEvent>> {
+  const result: Recorded<VehicleEvent> = await db.readEvent(device_id)
```
avatarneil: This seems unnecessary, why not just use `db.readEvent` inline instead of wrapping it in another function call?

mdurling: This was me originally. Trying to implement a thin service layer between the API and the DB. Mostly, this was because many DB methods throw when there are no records found and that was not the behavior I wanted. See `getVehicle` below for example.

packages/mds-db/schema.ts
```diff
@@ -106,6 +109,9 @@ const TABLE_COLUMNS: { [T in TABLE_NAME]: Readonly<COLUMN_NAME[]> } = {
     COLUMN.audit_event_type,
     COLUMN.audit_issue_code,
     COLUMN.audit_subject_id,
+    COLUMN.provider_event_id,
+    COLUMN.provider_event_type,
+    COLUMN.provider_event_type_reason,
```
mdurling, line 114: You are going to need to write a migration to add these columns to the `audit_events` table.

packages/mds-audit/api.ts
```diff
@@ -116,14 +119,27 @@ function api(app: express.Express): express.Express {
...
+        return next()
+      }
+      res.locals.audit_trip_id = audit_trip_id
+      res.locals.audit = await readAudit(audit_trip_id)
+      let { provider_id, provider_vehicle_id } = req.body
+      if ((!provider_id || !provider_vehicle_id) && res.locals.audit) {
+        provider_id = res.locals.audit.provider_id
+        provider_vehicle_id = res.locals.audit.provider_vehicle_id
+      }
+      const device = await readDeviceByVehicleId(provider_id, provider_vehicle_id)
```
mdurling: Shouldn't this be calling `getVehicle`? Also, there's a chance it won't be found so you probably check for `undefined` ... also, this middleware `app.use` is called for every request that starts with `/trips/:audit_trip_id` which is probably not what you want. I don't think you really need to do this here, or stash the event in res.locals.

janedotx:
@BigIdeas hey just fyi, we don't force push here.

packages/mds-db/schema.ts
janedotx: See this commit for how to do the migration btw: https://github.com/CityOfLosAngeles/mds-core/commit/bf241bff772b1efb2803c89f5e6721d63ed52d9b

packages/mds-audit/api.ts
ghost: My thoughts exactly...

packages/mds-db/schema.ts
ghost: Thanks for the example @janedotx - I added the db migration code, assuming I'll merge this PR after yours

ghost:
Thanks for the review everyone, comments should be addressed now

packages/mds-db/migration.ts
```diff
@@ -130,11 +134,24 @@ async function doMigration(exec: SqlExecuterFunction, migration: MIGRATION, migr
   }
 }
 
+async function alterAuditEventsColumns(exec: SqlExecuterFunction) {
```
mdurling, line 137: We are establishing a convention in one of Jane's PRs. Name this function `alterAuditEventsColumnsMigration` to make it more clear that you would not normally call this function outside of a migration.

packages/mds-db/migration.ts
```diff
@@ -130,11 +134,24 @@ async function doMigration(exec: SqlExecuterFunction, migration: MIGRATION, migr
...
+    `ALTER TABLE ${schema.TABLE.audit_events} ADD COLUMN ${schema.COLUMN.provider_event_type_reason} ${schema.COLUMN_TYPE.provider_event_type_reason}`
+  )
+}
+
 async function doMigrations(client: MDSPostgresClient) {
   const exec = SqlExecuter(client)
   // All migrations go here. createMigrationsTable will never actually run here as it is inserted when the
   // migrations table is created, but it is included as it provides a template for how to invoke them.
   await doMigration(exec, 'createMigrationsTable', async () => {})
+  await doMigration(exec, 'alterAuditEventsColumns', alterAuditEventsColumns)
```
mdurling, line 154: This becomes ` await doMigration(exec, 'alterAuditEventsColumns', alterAuditEventsColumnsMigration)`

mdurling (approved):
Minor nit, otherwise LGTM

avatarneil:
Please retarget this PR at @lacuna-tech/mds-core

PR #206, opened 2019-10-07
versioned values.yaml build output

oldschoolsysadmin:
Pull the version string out of each package.json, create a values.yaml file for subsequent use with helm


mdurling (approved):
LGTM ... do we need to update the `README` to outline when `yarn values` should be run?

jwtodd:
lgtm in in general with the following questions/statements:

as i understand it `./bin/mdsctl build` should still-do-the-right-thing, in that images will be built and tagged appropriately in the local repo

q: re the conditionally-push-to-image-repo pr [ https://github.com/CityOfLosAngeles/mds-core/blob/feature/jwtodd-conditionally-push-to-image-repo/bin/install#L212 ] what changes are required?

also, does `bin/mdsctl install:mds` require modification, ie point to the generated values.yaml?

oldschoolsysadmin:
> lgtm in in general with the following questions/statements:
> 
> as i understand it `./bin/mdsctl build` should still-do-the-right-thing, in that images will be built and tagged appropriately in the local repo

Correct.

> q: re the conditionally-push-to-image-repo pr [ https://github.com/CityOfLosAngeles/mds-core/blob/feature/jwtodd-conditionally-push-to-image-repo/bin/install#L212 ] what changes are required?

I'm not sure - this PR only creates the up-to-date values.yaml for the `apis:` map - shouldn't affect any part of the image build.

> also, does `bin/mdsctl install:mds` require modification, ie point to the generated values.yaml?

I'm not seeing where it points to any particular `values.yaml` file by default - are you referring to this?

```
    helm install ./helm/mds --name mds --namespace ${configs[namespace-mds]:-mds} \
      ${configs[values]:+$(helmOptions values ${configs[values]})} \
      ${configs[values-mds]:+$(helmOptions values ${configs[values-mds]})} \
```

PR #214, opened 2019-10-09
[mds-db] [mds-policy-author] [mds-policy] Automatically populate start_date, effective_date fields of a Geograp…

janedotx:
…hy when a Policy is published.

## PR Checklist

## Impacts
- [ ] Provider
- [ ] Agency
- [ ] Audit
- [ ] Policy
- [ ] Compliance
- [ ] Daily
- [ ] Native
- [ ] Policy Author

### DIFF
--- a/packages/mds-db/index.ts
+++ b/packages/mds-db/index.ts
@@ -1221,16 +1221,37 @@ async function deleteGeography(geography_id: UUID) {
   return geography_id
 }
 
-async function publishGeography(geography_id: UUID) {
+async function publishGeography(params: { geography_id: UUID; effective_date: Timestamp }) {
   try {
     const client = await getWriteableClient()
-    const geography = await readSingleGeography(geography_id)
+    const geography = await readSingleGeography(params.geography_id)
     if (!geography) {
       throw new NotFoundError('cannot publish nonexistent geography')
     }
-    const sql = `UPDATE ${schema.TABLE.geographies} SET read_only = TRUE where geography_id='${geography_id}'`
-    await client.query(sql)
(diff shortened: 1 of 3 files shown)


packages/mds-db/index.ts
```diff
@@ -1221,16 +1221,37 @@ async function deleteGeography(geography_id: UUID) {
...
+    // set effective_date to the provided param if it is earlier than what is already there
+    const vals = new SqlVals()
+    const conditions = []
+
+    if (!geography.read_only) {
+      conditions.push(`read_only = ${vals.add('t')}`)
+    }
+
+    if (!geography.publish_date) {
+      conditions.push(`publish_date = ${vals.add(now())}`)
```
owenoak, line 1241: It seems like there's a possibility that `now()` used here for the geo and `now()` used in the `publish_date` for the policy could drift slightly, meaning that the `publish_date` timestamps wouldn't match. ??  I suggest you pass `publish_date` in as well as `effective_date`.

packages/mds-db/index.ts
```diff
@@ -1221,16 +1221,37 @@ async function deleteGeography(geography_id: UUID) {
...
+    }
+
+    if (!geography.effective_date || geography.effective_date > params.effective_date) {
+      conditions.push(`effective_date = ${vals.add(params.effective_date)}`)
```
owenoak, line 1245: So this means that the policy might change after it's been marked `read_only`, which may violate Max's "immutable data never ever changes" policy.  From my perspective what you're doing is correct, but you may want to check with him.

PR #253, opened 2020-05-22
Catching Master up to make default

schnuerle:
Pushing dev to master so master can be the default branch in the repo, and new work can begin on dev.


jfh01 (approved):

marie-x (approved):

marie-x:
How are you going to manage those conflicts? How did those conflicts get in there?

schnuerle:
@karcass I think those conflicts are there because master was not the same as dev which was the starting point for the Lacuna work moving towards 0.1.  I assumed dev and master were the same before Lacuna forked dev, but it looks like that was not the case.

So there isn't a way to resolve the conflicts.  

I think what I'll have to do is make a new branch off of dev, delete master, then make the new dev copy the new master.

schnuerle:
Closing and trying the other option.

PR #255, opened 2020-06-15
Fix/alex/egress

oldschoolsysadmin:
This was causing an error on the helm install where we had an egress defined but no hosts.


oldschoolsysadmin:
oops

PR #266, opened 2021-01-19
Add docker registry push function to build

oldschoolsysadmin:
Since we can build images with the registry already present in their names, might as well add a function to push them as well.


oldschoolsysadmin:
oops

PR #267, opened 2021-01-30
[draft] better suffix than service

oldschoolsysadmin:
"Service" is resource type in Kubernetes.  Overloading the term for our microservices is going to lead to Helm yaml like this once we have sub-charts for microservice groups:

```
agtm-ingest:
  api:
    [...]
  service:
    serviceAccount: [...]
    service:
      annotations:
        [...]
```

The RPC pods serve as a backend for the API pods, so let's call them that.

Techniques used to create these changes:

```
rename 's/-service/-backend' $(find . -name "*-service")'
sed -i 's/-service/-backend/g $(find . -name '*.ts')
```


oldschoolsysadmin:
Sorry, wrong target

